PR #10, opened 2022-09-28
rc2nix: Add more *rc files

ncfavier:
Dolphin-related config.


pjones:
Thanks!

ncfavier:
There's also `ktrashrc` but that has a section that looks like `/home/user/.local/share/Trash` so maybe we want to replace `$HOME` with `${config.home.homeDirectory}` or something.

PR #24, opened 2023-10-19
implement hotkeys using KGlobalAccel application tab

### LINKED ISSUE #18: Migrate from KHotkeys to KGlobalAccel for custom command shortcuts
As explained [here](https://github.com/pjones/plasma-manager/issues/16#issuecomment-1546696208), KHotkeys has been deprecated in favor of KGlobalAccel. As such, I've done some investigative work to determine what the "Add Command..." button does in the Shortcuts system settings menu.

Creating a new command creates a new `.desktop` file in `~/.local/share/applications/`, containing the following:

```desktop
[Desktop Entry]
Exec=test
Name=test
NoDisplay=true
StartupNotify=false
Type=Application
X-KDE-GlobalAccel-CommandShortcut=true
```

This file is then referenced in `~/.config/kglobalshortcutsrc`:

```desktop
[test-2.desktop]
_k_friendly_name=test
_launch=none,none,test
```

I believe the `.desktop` file can be named anything, as long as it is unique.

References:
- https://invent.kde.org/frameworks/kglobalaccel/-/issues/3

nazarewk:
fixes https://github.com/pjones/plasma-manager/issues/18

implemented it in a slightly different manner than proposed by grouping all managed hotkeys under one application in configuration manager and having just one desktop entry with multiple actions.
Replicating "Add Command" flow (proposed in #18) resulted in terribly unreadable menu entry.

nix configurations is compatible with the master, I have simply added a new option `keys`.

looks like this:
Screenshot_20231019_155244
with following config
https://github.com/nazarewk-iac/nix-configs/blob/c4a5a18a0959671c0471acbd7ee46608b05b71d5/modules/profile/user/me/hm.nix#L62-L113


pjones:
This is fantastic!

Does this impact the `rc2nix.rb` script in any way?  Should it?

nazarewk:
To be honest this (2 days ago) is the first time I'm touching KDE so I'm taking baby steps by configuring most essential/annoying parts first. I did not take a look at rc2nix at all yet (probably never will) since I'm configuring Plasma from scratch without any prior config.
I basically got spoiled by config as a code of Sway over last 2 years  and want to replicate what's possible in plasma-manager, previously used XFCE for a decade.

toast003:
Did some testing and it works, although I have noticed that when you remove a hotkey from the home-manager config, it still shows up in the settings (which is not necessarily a problem, shortcuts do the same), however it stops working

magnouvean:
I updated this to work on the newest plasma-manager and it works great on my end! When using `overrideConfig` I didn't see any hotkeys still show up in settings (I didn't test if they were there when not using `overrideConfig` though). I'll merge this right away. Maybe there are some minor tweaks we could do, but this is miles better than the old solution anyway, which doesn't work at all anymore as far as I know.

Issue #33, opened 2023-12-10
. Escaping


hatch01:
I try to set my wallpaper through plasma-manager.
The corresponding config for me should be : 
```nix
      "plasma-org.kde.plasma.desktop-appletsrc"."Containments.1.Wallpaper.org.kde.potd.General"."FillMode" = 2;      
      "plasma-org.kde.plasma.desktop-appletsrc"."Containments.1.Wallpaper.org.kde.potd.General"."Provider" = "unsplash";
```
and this should produce something like this in `.config/plasma-org.kde.plasma.desktop-appletsrc` : 
```
[Containments][1][Wallpaper][org.kde.potd][General]
FillMode=2
Provider=unsplash
```

The problem is that the . In `org.kde.potd` are interpreted, and I did not find any method to escape them.
I don't have time yet to search and fix it myself. It is probably an issue in rc2nix. However, I am not sure because I am new to nix and plasma-manager.

PR #34, opened 2023-12-11
allow escaping .

hatch01:
As explained in #33, there are plenty of categories which contains dot in the name.
That's why I modified the files.nix to be able to escape dots if we like.


magnouvean:
Great work! LGTM

PR #37, opened 2023-12-19
Modules/kwin: add window titlebar button config

toast003:
Draft since since you can add any string you want even if it would generate a config that isn't valid

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,12 @@ Configuration is broken down into three layers:
            tooltipDelay = 5;
            theme = "breeze-dark";
            colorscheme = "BreezeDark";
-         }
+         };
+
+         kwin.titlebarButtons = {
+					left = [ "S" "F" ];
+          right = [ "H" "I" "A" "X" ];
+				};
          
          spectacle.shortcuts = {
            captureActiveWindow = "Meta+Print";
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@ Configuration is broken down into three layers:
          };
 
          kwin.titlebarButtons = {
-          left = [ "S" "F" ];
-          right = [ "H" "I" "A" "X" ];
+          left = [ "on-all-desktops" "keep-above-windows" ];
+          right = [ "help" "minimize" "maximize" "close" ]
         };
          
          spectacle.shortcuts = {
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ Configuration is broken down into three layers:
          };
 
          kwin.titlebarButtons = {
-					left = [ "S" "F" ];
+          left = [ "S" "F" ];
           right = [ "H" "I" "A" "X" ];
-				};
+        };
          
          spectacle.shortcuts = {
            captureActiveWindow = "Meta+Print";
--- a/modules/kwin.nix
+++ b/modules/kwin.nix
@@ -1,0 +1,43 @@
+{ config, lib, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.plasma;
+in
+{
+  options.programs.plasma.kwin = {
+    titlebarButtons.right = mkOption {
+      type = with types; nullOr (listOf str);
+      default = null;
+      example = [ "H" "I" "A" "X" ];
+      description = ''
+        Title bar buttons to be placed on the right.
+      '';
+    };
+    titlebarButtons.left = mkOption {
+      type = with types; nullOr (listOf str);
+      default = null;
+      example = [ "S" "F" ];
+      description = ''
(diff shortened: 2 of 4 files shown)


toast003:
@magnouvean should I leave it as is?

I'm thinking of either having something closer to what `kwinrc` actually has, but verifying that the options add buttons that exist ( so how it is rn but with checking):
``` nix
programs.plasma.kwin = {
  titlebarButtons.left = [ "S" "F" ];
  titlebarButtons.right = [ "H" "I" "A" "X" ];
};
```
or something that's longer to type, but that uses the same names as the system settings does:
``` nix
programs.plasma.kwin = {
  titlebarButtons.left = [ "on-all-desktops" "keep-above-windows" ];
  titlebarButtons.right = [ "help" "minimize" "maximize" "close" ];
};
```

bew:
I'd vote for longer names, so the config is self-documenting (no need for an additional doc to refer to)

magnouvean:
I do agree with @bew as this would be a higher-level so we may as well add an interface that makes more sense, as the default plasma config-file options aren't that helpful. Our aim for the higher-level interfaces should be closer to system-settings than config-files imo.

README.md
```diff
@@ -15,7 +15,12 @@ Configuration is broken down into three layers:
            tooltipDelay = 5;
            theme = "breeze-dark";
            colorscheme = "BreezeDark";
-         }
+         };
+
+         kwin.titlebarButtons = {
+					left = [ "S" "F" ];
```
magnouvean: The indenting here is a little off

toast003: The commit that changes this is supposed to be from https://github.com/pjones/plasma-manager/pull/36/ 😅
I'll change it when I get home

toast003: Got mixed them up, it's fixed up now

toast003:
Got it, long names it is!

toast003:
Spent quite a while working on it, but I can't quite get it to work ( it doesn't respect the order you specify the buttons to be in)
I'll give it another look tomorrow

I didn't want to push know-bad code here but I do need help for this so I did a patch with my changes 
[wip.zip](https://github.com/pjones/plasma-manager/files/13720970/wip.zip)

magnouvean:
Right, I'll try to take a look it later today if you haven't figured it out by then :)

toast003:
Figured it out, I don't think it's the cleanest way of doing it but it works

modules/kwin.nix
```diff
@@ -0,0 +1,86 @@
...
+
+  # Gets the index and returns the short name in that position
+  getShortNameFromIndex = position: builtins.elemAt validTitlebarButtons.shortNames (strings.toInt position);
+in
+{
+  options.programs.plasma.kwin = {
+    titlebarButtons.right = mkOption {
+      type = with types; nullOr (listOf (enum validTitlebarButtons.longNames));
+      default = null;
+      example = [ "H" "I" "A" "X" ];
```
magnouvean: Shouldn't we use the long names here

modules/kwin.nix
```diff
@@ -0,0 +1,86 @@
...
+      description = ''
+        Title bar buttons to be placed on the right.
+      '';
+    };
+    titlebarButtons.left = mkOption {
+      type = with types; nullOr (listOf (enum validTitlebarButtons.longNames));
+      default = null;
+      example = [ "S" "F" ];
```
magnouvean: And here

modules/kwin.nix
```diff
@@ -0,0 +1,86 @@
...
+      "N"
+      "S"
+      "I"
+      "A"
+      "X"
+      "H"
+      "L"
+      "B"
+      "F"
+    ];
```
magnouvean, line 31: Here it might have been better using a set to get translations between the long and short names. If it works as it should though it probably is fine for now as we would then need to change other parts of the code though. I will see if there is a quick fix.

magnouvean (commented):
Looks good for the most part I reckon :)

modules/kwin.nix
magnouvean: In the example attribute I am referring to here

toast003: Very much so, I just overlooked it

modules/kwin.nix
toast003: Main reason I used a list is cause I already needed a list of long names for the option types so that way I don't repeat stuff, although I would be open to using sets as you said

magnouvean: Ah OK. I missed that. Then this is a good solution I reckon.

magnouvean (approved):
Looks good now

Issue #35, opened 2023-12-17
Escaping `[]`


haizaar:
Good day,

I have the following configuration I want to apply under `programs.plasma.configFile`:
```nix
        "kwinrc"."Wayland"."InputMethod[$e]" = "/run/current-system/sw/share/applications/com.github.maliit.keyboard.desktop";
```

which results in the following broken configuration however in `kwinrc`
```desktop
[Wayland]
InputMethod\x5b$e\x5d=/run/current-system/sw/share/applications/com.github.maliit.keyboard.desktop
```

Is there a way to escape `[]` or is it a plasma-manager bug?

magnouvean:
I am pretty sure this is a kwriteconfig bug, which is what we are using at the moment as this is bundled with kde. We probably should change this method though as this tool is quite limited. I assume that if you run:
```
kwriteconfig5 --file kwinrc --group Wayland --key 'InputMethod[$e]' "/run/current-system/sw/share/applications/com.github.maliit.keyboard.desktop"
```
or
```
kwriteconfig5 --file kwinrc --group Wayland --key 'InputMethod\[$e\]' "/run/current-system/sw/share/applications/com.github.maliit.keyboard.desktop"
```
you get similar results? If this doesn't work it is a kwriteconfig issue (and also a plasma-manager issue). Kwriteconfig is quite limited in this regard. We probably should write our own little tool for this with some simple python-script or something.

toast003:
You can do this and I'll bring you closer
```
programs.plasma = {
	enable = true;
	configFile = {
		"kwinrc" = {
			"Wayland" = {
				"InputMethod" = "/run/current-system/sw/share/applications/com.github.maliit.keyboard.desktop";
				configGroupNesting = [ "Wayland" "$e" ];
			};
		};
	};
};
```
However, even if the script does do everything just fine kwriteconfig seemss to be bugged since it writes this to kwinrc
```
[Wayland][\x24e]
InputMethod=/run/current-system/sw/share/applications/com.github.maliit.keyboard.desktop
```

haizaar:
Thank you for the quick response guys. After playing with it further I found out that the `[$e]` part is not essential - the configuration works find without it, at list in my particular use case.

I think it's still a bug (somewhere down the chain) but at there is a workaround, sort of.

toast003:
We could be kinda hacky and just replace \x24 with $ after running kwriteconfig, but only as temporary thing until we do #1

magnouvean:
[...]

PR #39, opened 2023-12-22
Replace kwriteconfig5 with custom script

magnouvean:
This is a big change so it will probably need some testing. This replaces kwriteconfig5 with a custom-built script written in python using the built-in configparser, which in this case I think does the job fine. Therefore no python-dependencies needed. I have tested it on my own configs and cant see any obvious bugs and the config-files look fine to me, but I am sure there are some limitations I haven't seen.

By introducing our own config-writer we get more functionality and are able to combat some weird limitations in kwriteconfig5, and we also are able to run everything in one command, which probably makes this a bit faster.


toast003:
I don't use hotkeys so not sure if that works, but so far everything seems to work just fine

toast003:
Also pretty sure this fixes #35, I'll test it out

Edit: It does!
Both with the original configuration @haizaar had and with [my comment](https://github.com/pjones/plasma-manager/issues/35#issuecomment-1860417087)

magnouvean:
Yes this should in general be an improvement in many ways given hov limited kwriteconfig is, including the [$e] thing, which can in many ways be useful in configs to ensure that settings aren't overwritten. My config is also quite limited so I probably will test this out a little more in the coming days in a vm or something, and if no obvious problems are found by then I reckon we'll just merge this :)

magnouvean:
An update on this. I have hit a bit of a road-block as I found that some of the apps (konsole is the one I found so far), can have keys not in any category/group (just placed at the top). A.t.m this will cause a crash. I tried using PyQt5 instead of configparser, but that really didn't help (it doesn't seem to be the same format as the kde configuration files). There also doesn't really seem to be many well-maintained kconfig file-parsers/writers, so using configparser is probably what we should use (initially anyway, but eventually replacing it shouldn't be much work). As it is now configparser doesn't natively support keys outside any group, so I will probably have to come up with some clever code to add support for this.

haizaar:
Looks like a common limitation of the ConfigParser.

My 2c:

You can prepend a dummy section name at the top of the file, parse/modify,
and then strip that top section off from the resulting string before
writing it on disk.

You can use a long, convoluted name for the dummy section to minimise the
chance for collisions, e.g. "dummy_bcf036b6f33e182d4705f4f5b1af13ac"

magnouvean:
Yeah that resembles what I'll be trying to do I reckon. Maybe even just do the parsing manually given how simple ini files really are.

magnouvean:
Having gotten increasingly frustrated with `configparser`, I ended up just writing my own parser, which was quite easy given how simple the config files are syntactically. It now works just as it should on my config, but my config is probably one of the simpler ones so there could well be some bugs left. I have tested though that nested sections work and so does [$e] in keys (hence the issue mentioned is still solved). The issue mentioned in my previous comment with `konsolerc` is also fixed last I tried it. The script also runs quite quick as far as I can tell. If anyone wants to test feel free to do so (though I advise to take backups first) and see if it works. I will probably do some more testing as well before merging, as this is quite a big change, but if no major bugs are encountered after a bit of testing I reckon we'll just merge this, as it at least is a better solution than using `kwriteconfig5`. It also should be easy to eventually move towards using some dbus thing if that eventually is desired, as the only thing we'll have to rewrite is the `KConfParser` class :)

magnouvean:
Right, after a little bit more testing, I am pretty happy with this. Everything I tested it with worked just fine including hotkeys as far as I can see.

PR #68, opened 2024-02-10
Various bug fixes

### LINKED ISSUE #25: Shortcuts with multiple key combinations are set incorrectly
Making meta+ctrl+up and meta+ctrl+right switch one desktop to the right though the settings writes `Switch One Desktop to the Right=Meta+Ctrl+Up\tMeta+Ctrl+Right,Meta+Ctrl+Right,Switch One Desktop to the Right`  in .config/kglobalshortcutsrc, while doing the same thing with plasma manager with 
```
shortcuts = {
	"kwin" = {
		"Switch One Desktop to the Left" = [ "Meta+Ctrl+Left" ];
		"Switch One Desktop to the Right" = [ "Meta+Ctrl+Right" "Meta+Ctrl+Up"];
	};
};
```
writes `Switch One Desktop to the Right=Meta+Ctrl+Right\tMeta+Ctrl+Up,,`, which doesn't work

magnouvean:
This PR addresses some bugs I, and others in the issues, have encountered. Also this does some minor refactoring. The bugs this addresses in particular is:
- Adds functionality for multiple keybindings (which should close #25 )
- Fixes problems with escaping non-ASCII characters (like ¤), and now only escapes \n and \t
- Removes check for `file` `dataFile` and `configFile` being empty before applying the config-writing script, as this seems redundant now with the new config-writing system (I am not really sure why it was there in the first place)

Specifically the last solution I would be welcome to thoughts on, as I am not sure if I see if there was ever a good reason for checking `(builtins.length (builtins.attrNames cfg) > 0)` before running the config-writing part, but I could be missing something. The config-writing system should be able to handle empty inputs, so I don't think it should be a problem. I have tested that plasma-manager works as expected with this change and the bugs should be fixed.

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -91,7 +91,7 @@ let
       # last_run and is located in $XDG_DATA_HOME/plasma-manager. When we
       # reset all the other config-files these startup-scripts should be
       # re-ran, so we delete these files to ensure they are.
-      ++ [ "for file in ${config.xdg.dataHome}/plasma-manager/last_run*; do ${removeFileIfExistsCmd "$file"}; done" ]));
+      ++ [ "for file in ${config.xdg.dataHome}/plasma-manager/last_run_*; do ${removeFileIfExistsCmd "$file"}; done" ]));
 in
 {
   options.programs.plasma = {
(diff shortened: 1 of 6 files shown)


toast003 (commented):
Looks good to me, but I don't have the time to test out things right now

toast003:
> Specifically the last solution I would be welcome to thoughts on, as I am not sure if I see if there was ever a good reason for checking `(builtins.length (builtins.attrNames cfg) > 0)` before running the config-writing part, but I could be missing something. The config-writing system should be able to handle empty inputs, so I don't think it should be a problem. I have tested that plasma-manager works as expected with this change and the bugs should be fixed.

That might be cause 'kwriteconfig5' didn't handle empty inputs properly if I have to guess

magnouvean:
That was what I though as well. Great, thanks for the input, I'll merge then right away as everything works on my side :)

PR #70, opened 2024-02-21
add github CI

m1-s:
I just started using this project and noticed potential for improvement. This is the first of a series of PRs. This PR adds a github action file that executes `nix flake check` for the main flake and the example flake. This workflow is automatically executed on every PR.

:warning: The maintainers of this repository may have to enable actions in the repository settings for this to work.

I tested this in a PR in my own fork https://github.com/m1-s/plasma-manager/pull/1/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,17 @@
+name: "Check"
+on:
+  pull_request:
+# cancel previous runs when pushing new changes
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v22
+      with:
+        extra_nix_config: "system-features = kvm nixos-test"
+    - run: nix flake check -L --keep-going
+    - run: nix flake check -L --keep-going ./example
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
-/example/result
+result-*
+result
+.direnv
--- a/example/home.nix
+++ b/example/home.nix
@@ -1,4 +1,4 @@
-{ ... }:
+{ pkgs, ... }:
 {
   programs.plasma = {
     enable = true;
--- a/script/write_config.py
+++ b/script/write_config.py
@@ -3,7 +3,7 @@ import json
 import os
 import re
 import sys
-from typing import Dict
+from typing import Dict, Optional
 
 
 class KConfParser:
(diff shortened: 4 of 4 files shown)


magnouvean:
Looks pretty good, nice work, and something I think would be welcome for the project. Only thing I am not sure about is if  `Optional[str]` is a lot better than `str | None`. From what I have gathered, in newer python-versions `str | None` is built in and if anything is the more "modern" way of typesetting optionals.

m1-s:
> Looks pretty good, nice work, and something I think would be welcome for the project. Only thing I am not sure about is if `Optional[str]` is a lot better than `str | None`. From what I have gathered, in newer python-versions `str | None` is built in and if anything is the more "modern" way of typesetting optionals.

The only reason I made that change was because the python tooling complained about it. Can you clarify what you expect me to do here? I did not find anything about Optional being "built in".

magnouvean:
I see. Not sure why the tooling would be complaining about it tbh as it to me seems like `str | None` is the "new" way of doing this kind of typing while Optional is more the old way of doing this (see: https://docs.python.org/3/library/typing.html#typing.Optional). In the end it doesn't really matter which of them we choose, they are equivalent after all, but I don't really think it needs changing at all.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v22
+      with:
+        extra_nix_config: "system-features = kvm nixos-test"
+    - run: nix flake check -L --keep-going
+    - run: nix flake check -L --keep-going ./example
```
magnouvean, line 17: Why is `./example` being checked separately? On my end running `nix flake check` also does check `./example` it seems like.

m1-s: No these are 2 separate flakes. They need to be checked individually.

m1-s:
> I see. Not sure why the tooling would be complaining about it tbh as it to me seems like `str | None` is the "new" way of doing this kind of typing while Optional is more the old way of doing this (see: https://docs.python.org/3/library/typing.html#typing.Optional). In the end it doesn't really matter which of them we choose, they are equivalent after all, but I don't really think it needs changing at all.

Not sure why the tooling complained either. I created https://github.com/NixOS/nixpkgs/pull/290747 which may be related.

magnouvean:
OK, then this looks good to me. I reckon the python-typing thing should stay unchanged though so if you can revert this then I'll merge as soon as possible :). Good work!

m1-s:
It wont build without this change.

magnouvean:
All right then, we can just change it, it really doesn't matter. I'll merge when I get home in a few hours

PR #71, opened 2024-02-21
bump to 23.11

m1-s:
Bump to current nixpkgs stable 23.11


magnouvean:
Nice, seems to work well. I'll merge right away

PR #72, opened 2024-02-21
refactor basic test

m1-s:
This PR replaces the `pkgs.nixosTest` call with `pkgs.testers.nixosTest` which is the new better nixos test. Furthermore, we are calling the test via `callPackage` now.

### DIFF
--- a/test/basic.nix
+++ b/test/basic.nix
@@ -1,12 +1,10 @@
-{ pkgs, home-manager, module }:
-
+{ testers, home-manager-module, plasma-module, writeShellScriptBin, libsForQt5 }:
 let
-  script = pkgs.writeShellScriptBin "plasma-basic-test" ''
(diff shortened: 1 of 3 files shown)


test/basic.nix
```diff
@@ -24,25 +22,27 @@ let
...
-    inherit module home-manager homeConfig;
-  };
-in
-pkgs.nixosTest {
-  name = "plasma-basic";
-
-  nodes.machine = { ... }: {
-    imports = [ user ];
+    home-manager.users.fake = {
+      home.stateVersion = "21.11";#"23.11";
```
magnouvean: Should this just be 23.11?

m1-s: yes. it is now.

magnouvean:
Nice, great work!

Issue #62, opened 2024-01-24
Problems with plasma 6


magnouvean:
This is a tracker issue for things not working in plasma-manager on the upcoming kde plasma 6 version. If there are any bugs/configuration-problems in plasma-manager, feel free to post them here. At the moment kde plasma isn't available for nixos, but it is available for other distributions to test some of the tools/configurations plasma-manager to find potential problems. As already noted in #59 @mlyxshi, the way we generate panels may be broken (hopefully this is just a problem with the qdbus command we send in and hence it hopefully is an easy fix).

magnouvean:
I set up my own virtual machine with kde neon and tested some things. It appears that the panel-approach works just fine without changes on kde plasma 6 (as far as I saw), that is after installing some package (I believe it was called `qtchooser`) so that `qdbus` was available (this wasn't installed by default unlike in nix). I also tested that the various plasma command-line tools we use (like `plasma-apply-colorscheme`) are included and continue to work in this kde plasma version. Overall I think we should for the most part be good with little change in plasma-manager needed (though we will perhaps need to adjust the name of some keys or some values written to config-files, although this does look very similar, and I haven't seen anything so far that needs adjustment).

Other than that I will say that plasma 6 does look very good at first look, so this is something to look forward to!

mlyxshi:
[kde6](https://github.com/nixos/nixpkgs/pull/286522) is available in nixos-unstable. The only problem is the new default floating panel style (I prefer the old non-floating style). Could you please support this boolean option?

magnouvean:
Yeah, we'll have to look into that. Hopefully this capability has been added to plasma desktop scripts as well, although I'm not seeing anything obvious in the documentation for it yet. I also see that the maxwidth stuff doesn't work either. On a positive note it appears that the panels at least are created on the correct screen, which was always a bit irritating when using multi-monitors :).

mailbun:
Plasma6 added some new global shortcuts for Spectacle, so it would be appreciated if new options in spectacle.nix were added.

[...]

PR #77, opened 2024-02-29
Support kde6 panel floating

mlyxshi:
In `plasma-interactiveconsole`, I try to print all object key of panel and find this key `floating`
#62

### DIFF
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -67,6 +67,7 @@ let
         example = "autohide";
         description = "The hiding mode of the panel.";
       };
+      floating = lib.mkEnableOption "enable or disable floating style";
       widgets = lib.mkOption {
         type = with lib.types; listOf (either str widgetType);
         default = [
@@ -143,6 +144,7 @@ let
 
     var panel = new Panel;
     panel.height = ${builtins.toString panel.height}
+    panel.floating = ${lib.boolToString panel.floating}
     var panelWidgets = {}
     ${if panel.alignment != null then "panel.alignment = \"${panel.alignment}\"" else ""}
     ${if panel.hiding != null then "panel.hiding = \"${panel.hiding}\"" else ""}
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -67,7 +67,7 @@ let
         example = "autohide";
         description = "The hiding mode of the panel.";
       };
-      floating = lib.mkEnableOption "enable or disable floating style";
+      floating = lib.mkEnableOption "Enable or disable floating style (plasma 6 only).";
       widgets = lib.mkOption {
         type = with lib.types; listOf (either str widgetType);
         default = [


toast003:
Thanks for finding this out @mlyxshi <3

It's not in the [documentation](https://develop.kde.org/docs/plasma/scripting/api/) yet

toast003:
It doesn't work in plasma 5, so we should either say that it only works on 6, or set the panel to be floating like we do now if you're running plasma 5
Might take a look at it tomorrow

mlyxshi:
> It doesn't work in plasma 5, so we should either say that it only works on 6, or set the panel to be floating like we do now if you're running plasma 5

It is a little bit tricky. Floating is one of the location in kde5, but a boolean key in kde6.

I don't know how to handle this transition. Personally, I have upgraded all my linux desktop to kde6 except steamdeck. This pr doesn't change the current kde panel configuration in my steamdeck, so I think it's backward compatible.

toast003:
It is backwards compatible in the sense that it doesn't break anything in plasma 5, however I think that having two different ways to set a panel to floating is confusing.

Imo we should check if we're running on plasma 5 or 6, and then change `location` or `floating` depending on the version

toast003:
@mlyxshi (or anyone that's running plasma6) can you run `print(scriptingVersion)` in `plasma-interactiveconsole`?

mlyxshi:
> print(scriptingVersion)

20

>  I think that having two different ways to set a panel to floating is confusing.

It's originated from plasma... I'm also confused

toast003:
Plasma 5 also has 20 so I'll have to use `applicationVersion` instead

>It's originated from plasma... I'm also confused

My bad, I don't think I'm explaining myself properly :sweat_smile: 

Before plasma 6, if you wanted to make a floating panel you needed to set `location` to `floating`. Now there's a dedicated property for that ( the `floating` key you found).

If we add an option for `floating` just like you have here we could make a floating panel both by setting `location` to floating, and by setting `floating` to true.

What I think we should do remove the ability to set `location` to floating, and keep the floating option, and then use the key you found if you're running plasma 6, or the old way if not

modules/panels.nix
```diff
@@ -67,6 +67,7 @@ let
         example = "autohide";
         description = "The hiding mode of the panel.";
       };
+      floating = lib.mkEnableOption "enable or disable floating style";
```
magnouvean: Maybe add a disclaimer that it is plasma 6 only. Should also start with a capital letter and end with period as the rest of the descriptions

magnouvean (commented):
Looks good to me otherwise.

toast003:
Yup same for me
We can implement what I said in a separate pr

magnouvean:
One thing though I am not sure about is how we would determine if you are running plasma 5 or 6. Would it not be possible to have both installed at the same time? And even if it isn't would it be possible for plasma-manager which is built on top of home-manager to be able to check if plasma 6 or 5 is installed given that they are declared on the system-level

toast003:
https://develop.kde.org/docs/plasma/scripting/api/#version-numbers

>Starting with KDE SC 4.5, the version number of both the scripting API and the application is available to the script via the >following read-only properties:
>
>    `string applicationVersion`: the version of the application, e.g. 5.20.3
>    `number scriptingVersion`: the version of the scripting API; e.g. for Plasma 5.20 it is 20.

`scriptingVersion` is still 20 on plasma 6, but `applicationVersion` does change

magnouvean:
Right, forgot that we can check for this in the script itself and not from nix. Sound like a good idea then!

mlyxshi:
> Right, forgot that we can check for this in the script itself and not from nix. Sound like a good idea then!

Sounds great. I don't really want to go back to plasma5 and test floating stuff, because I don't use it anymore.
Another PR is welcome.

magnouvean:
I understand that, tbh I am not overly keen to dive back and forth either, but will probably do some minimal testing to keep plasma 5 support somewhat alive. I reckon we can merge as soon as the above conversation about the description is resolved I reckon. The description could be for instance `Enable or disable floating style (plasma 6 only).`. We could just remove the plasma 6 only part when we implement plasma 5 support for the same option.

mlyxshi:
@magnouvean   description updated

PR #80, opened 2024-03-08
Add dodgewindows to hiding for plasma 6

WCollier:
Plasma 6 includes a `dodgewindows` option for panels. I'm not sure if backwards compatibility is a concern here. However, this seemingly isn't a concern with the `floating` option.


magnouvean:
Nice spot, thanks. I'll merge right away!

PR #86, opened 2024-03-17
Use nested attrsets for nested sections

MatthewCash:
Right now, in plasma-manager, to produce a nested section such as
```ini
[section1][section2]
key=value
```
you would set
```nix
configFile.something."section1.section2".key = "value";
```
with the two sections in a single string separated by "."

This leads to two unfortunate side effects:

1. Section names that contain "." need a workaround like the `configGroupNesting` key to explicitly set the section name.
2. The entire group name must be specified for every key like this:
```nix
"plasma-org.kde.plasma.desktop-appletsrc" = {
  "Containments.2.Applets.1"."plugin" = "org.kde.plasma.kickoff";
  "Containments.2.Applets.1.Configuration"."PreloadWeight" = 100;
  "Containments.2.Applets.1.Configuration"."popupHeight" = 511;
  "Containments.2.Applets.1.Configuration"."popupWidth" = 671;
  "Containments.2.Applets.1.Configuration.General"."favoritesPortedToKAstats" = true;
  "Containments.2.Applets.1.Configuration.General"."icon" = "nix-snowflake-white";
  "Containments.2.Applets.1.Configuration.General"."systemFavorites" = "suspend,hibernate,reboot,shutdown";
};
```

This PR changes how sections are described by using nested attrsets for nested sections. This is done by "flattening" the nested attrset into a single string separated with "][". A section with a "." can now be written literally without `configGroupNesting`, and the nested section described previously can be defined in a more readable format:

```nix
"plasma-org.kde.plasma.desktop-appletsrc".Containments."2".Applets."1" = {
  plugin = "org.kde.plasma.kickoff";
  Configuration = {
    PreloadWeight = 100;
    popupHeight = 511;
    popupWidth = 671;
    General = {
      favoritesPortedToKAstats = true;
      icon = "nix-snowflake-white";
      systemFavorites = "suspend,hibernate,reboot,shutdown";
    };
  };
};
```

This is a breaking change for anyone who uses nested sections or `configGroupNesting`, but I think it is worthwhile to consider since, in my opinion, the API feels more natural and more like what users would expect. Maybe this new behavior could be opt-in?


magnouvean:
Very nice. I'll need to do some testing next week . This will break all the existing configs so we probably should make this clear to the users in some way. Looks very well implemented though, and this is a much better solution than the current existing one!

magnouvean:
I have tested this on my config and tried a lot of different stuff, and this seems to work great. Very well done! I'll merge this right away I think and pin an issue notifying people that this is what will be breaking their configs (and how to fix it).

Issue #87, opened 2024-03-18
Configure kate with plasma-manager


Asqiir:
A while ago, [I started configuring kate with plasma-manager](https://github.com/pjones/plasma-manager/discussions/63#discussioncomment-8423722). I think it is worthwile to put my results into this project in the form of a handful new mid-level configurations.

Kate is the text editor from the KDE project and as such, its config files work in the same way as most KDE config files. This specifically implies "these are stateful config files and as such belong into the users home-environment". That is why I believe this project to be the right place for the effort.

I want to implement mid-level (high-level?) options for the following behaviour. What do you think?

* [x] indentation (tabs, spaces, how wide?)
* [ ] line breaks (forced/none, at what width, \n vs \r\n)
* [x] The editor theme. How should the configuration react when a theme with the same filename already exist (thus, an existing file could be destroyed) or a theme with the same name (a field within the file) already exists (thus, undefined behaviour in kate could occur)?)
* [ ] Which menus and tools to show by default, which key shortcuts to focus/unfocus/show/hide.
* [ ] Font type, size, line-height, ... The main open question here would be: should the configuration fail if the selected font cannot be found?

magnouvean:
Any new modules will always be welcome! What I would suggest is to make options which uses plasma-manager's `configFile`, `dataFile` and so on for configuration. I don't think you really need to worry about if a file already exists with the same name as adding configurations using `configFile` and `dataFile` doesn't remove already existing files, just adds new keys within the files (doing different actions based on the present files isn't really encouraged or well supported in nix anyway afaik). For the font I don't think you really need to worry about that either, that can be up to the user to make sure they have the font installed (checking this in nix is not easy either).

Also I would suggest that the module should be placed under `modules/apps/kate.nix` and use `programs.kate` prefix, just like in #69. Feel free to look for inspiration in that pr, as this will be quite similar in many ways I would assume.

Asqiir:
@magnouvean 
When mentioning "files that might be overriden" I am talking about custom editor theme files. This is a separate file, in a format distinct from usual KDE config files.

My main points are:
* this custom format seems not to work with `config.plasma.dataFile`
* I expect users to always replace this file with an entire new file

As long as we expect users to stick to the kde naming scheme (`<themename>.theme`), and them deleting the theme from original directory everything should be fine. But is that a reasonable assumption?

# Some more detailed information

[Documentation from KDE](https://api.kde.org/frameworks/syntax-highlighting/html/)
[...]

PR #88, opened 2024-03-19
add indentation options to kate

Asqiir:
This is the first contribution to #87 .
It allows users to configure the way kate handles indentation.

# Reasoning about decisions:
I  decided to put all options into `config.programs.kate.editor`. The `editor` option should contain all options that act on configuration into the rectangle where users type text in (as opposed to: the console, the `project` tab, the menus and sidebards).

# Known limitations:
* `kate` needs to be restart after switching the configuration. I could not figure out how to make `kate` do a hot reload on the changed `katerc`.
* I would like to have the message in the assertion reference the variables in question not hard-coded by their name. I would prefer if the compilation fails if someone was to change the names of the variable but forgot to update the message.

Do you want me to put all changes into a single MR (leaving this one open), or do you prefer each change to be its own PR?


Asqiir:
Should we add the `katerc` (`.config/katerc`) into the `defaultResetFiles` in [`modules/files.nix`](https://github.com/pjones/plasma-manager/blob/trunk/modules/files.nix)? (I'm not sure what that list does.)

modules/apps/kate.nix
```diff
@@ -0,0 +1,89 @@
...
+  # compute kate's magic TabHandlingMode
+  boolToInt = b: if b then 1 else 0;
+  t = boolToInt config.programs.kate.editor.indent.tabFromEverywhere;
+  s = boolToInt config.programs.kate.editor.indent.undoByShiftTab;
+  tabHandlingMode = t + 2 * s - 1;
+in
+{
+  options.programs.kate.editor = {
+    tabWidth = lib.mkOption {
+      description = "The width of a single tab (\\t) sign (in number of spaces).";
```
magnouvean: Shouldn't it just be \t here and not \\\t?

modules/apps/kate.nix
```diff
@@ -0,0 +1,89 @@
...
+      default = true;
+      type = lib.types.bool;
+    };
+  };
+
+  config.assertions = [
+    {
+      assertion = tabHandlingMode >= -1;
+      message = "Either 'undeByShiftTab' or 'tabFromEverywhere' needs to be enabled!";
+      # kate does not seem to support disabling both.
```
magnouvean: I would put the comment above the message. Maybe move it up and rewrite as something like "kate doesn't seem to support disabling undeByShiftTab and tabFromEverywhere at the same time"

magnouvean (commented):
See my two small comments. Looks good to me otherwise assuming you have tested that the module works (I don't really use kate so I can't really test)

modules/apps/kate.nix
Asqiir: I changed the error message.

modules/apps/kate.nix
Asqiir: `\t` is rendered to the tab whitespace.
`\''t` is rendered to the tab character. I switched to that.

Source: [the nix documentation](https://nixos.org/manual/nix/stable/language/values.html#type-string)

Asqiir:
~~found bugs :( 
(do not merge yet, I hope to fix it today)~~

fixed

magnouvean:
I added an enable option (which is in general good practice) and assigned `config.programs.kate` to a variable `cfg` (which we do throughout the project to shorten things up a bit). See my two earlier commits. Are you happy enough with this so that we can merge this now @Asqiir?

Asqiir:
@magnouvean we can merge this imo

PR #95, opened 2024-03-25
Add colortheme configuration to kate

Asqiir:
This is the second contribution to #87.

This contribution adds support for kate's colorthemes. A user can either pick one of the system themes by name, or a file for a custom theme.

The custom file will never override any existing file, nor will it introduce name collisions.

# Known limitations

When picking a theme by name there is no check whether such a theme exists. The existing themes live in the `ksyntaxhighlighting` framework in `data/themes`.
But these existing themes do not live in any of `${XDG_DATA_DIRS}`. They are included as a linked library. I don't know if there is any non-ugly way to access these (option 1: via an overlay. option 2: going through multiple layers of built-inputs).

As an artefact of that process (when I tried to find the themes) there is an option `home.kate.package` that allows the user to pick a package to be installed. By default, this is `pkgs.kate`. That changes the default behaviour of `programs.kate.enable`: When enabling this option, a package will be installed. I believe this is a more natural behaviour for the enable option.


modules/apps/check-theme-name-free.sh
magnouvean, line 1: I think there is a better directory for this file. You could change the name to something like "kate-check-free-theme-name.sh" and move it to the `script` folder. Alternatively you could also add a separate folder for the kate module, move the `kate.nix` file to this folder with the name of `default.nix` and move this script in this directory as well (you would maybe have to adjust the import in `apps/default.nix` as well though). I think the latter option is the most common way in `nixpkgs`, but I think both options are fine. This would be a lot more organized than just leaving the script in the apps directory.

modules/apps/kate.nix
```diff
@@ -96,6 +104,56 @@ in
...
+  config = {
+    home.packages = lib.mkIf (cfg.enable && cfg.package != null) [ cfg.package ];
+
+    # In case of using a custom theme, check that there is no name collision
+    home.activation.checkKateTheme = lib.mkIf (cfg.enable && cfg.editor.theme.src != null) (lib.hm.dag.entryBefore [ "writeBoundary" ]
+    ''
+      ${./check-theme-name-free.sh} ${cfg.editor.theme.name} ${pkgs.jq}/bin/jq
+    '');
+
+    # In case of using a system theme, there should be a check that there exists such a theme
```
magnouvean: Is there something missing here?

modules/apps/kate.nix
```diff
@@ -96,6 +104,56 @@ in
...
+    target = "org.kde.syntax-highlighting/themes/${cfg.editor.theme.name}.theme";
+  };
+
+  config = {
+    home.packages = lib.mkIf (cfg.enable && cfg.package != null) [ cfg.package ];
+
+    # In case of using a custom theme, check that there is no name collision
+    home.activation.checkKateTheme = lib.mkIf (cfg.enable && cfg.editor.theme.src != null) (lib.hm.dag.entryBefore [ "writeBoundary" ]
+    ''
+      ${./check-theme-name-free.sh} ${cfg.editor.theme.name} ${pkgs.jq}/bin/jq
```
magnouvean: Here I would suggest creating a nix-derivation here which runs the script. That way you don't have to send in the jquery path as a command-line argument, and in general is a better practice. See [here](https://github.com/pjones/plasma-manager/blob/d0e5698aa41e331ab36e9700b4be51f40127a47d/lib/writeconfig.nix#L4) for how we did with the config-writing script

modules/apps/kate.nix
```diff
@@ -96,6 +104,56 @@ in
...
     };
   };
+
+
+  # ==================================
+  #     COLORTHEME
+
+  options.programs.kate.editor.theme = {
+    src = lib.mkOption {
+      description = "The path of a theme file for the KDE editor (not the window color scheme). Obtain a custom one by using the GUI settings in kate. If you want to use a system-wide editor color scheme set this path to null. If you set the metadata.name entry in the file to a value that matches the name of a system-wide color scheme undesired behaviour may occur. The activation will fail if a theme with the filename `<name of your theme>.theme` already exists.";
```
magnouvean: Move over multiple lines here as well.

modules/apps/kate.nix
```diff
@@ -20,6 +19,15 @@ in
     enable = lib.mkEnableOption ''
       Enable configuration management for kate.
     '';
+
+    package = lib.mkPackageOption pkgs "kate" {
+      default = [ "kate" ];
+      example = "pkgs.libsForQt5.kate";
+      extraDescription = "Which kate package to install. Use `pkgs.libsForQt5.kate` in Plasma5 and `pkgs.kdePackages.kate` in Plasma6. Use `null` if home-manager should not install kate (use this if you want to manage the settings of this user of a system-wide kate installation).";
```
magnouvean: Move comment over multiple lines

modules/apps/kate.nix
```diff
@@ -96,6 +104,56 @@ in
...
+
+  options.programs.kate.editor.theme = {
+    src = lib.mkOption {
+      description = "The path of a theme file for the KDE editor (not the window color scheme). Obtain a custom one by using the GUI settings in kate. If you want to use a system-wide editor color scheme set this path to null. If you set the metadata.name entry in the file to a value that matches the name of a system-wide color scheme undesired behaviour may occur. The activation will fail if a theme with the filename `<name of your theme>.theme` already exists.";
+      type = lib.types.nullOr lib.types.path;
+      default = null;
+    };
+
+    name = lib.mkOption {
+      description = "The name of the theme in use. May be a system theme. If a theme file was submitted this setting will be set automatically.";
```
magnouvean: Multiple lines here also

magnouvean (commented):
See my comments for some initial possible tweaks

modules/apps/kate.nix
Asqiir: tldr: No, this is a note that there should be a check – if I just knew how to. (See "known limitations" above.) I improved the comment.

long form: I would like to check that a user can only select valid themes – that are those already packaged in the installed kate package. I could not figure out a viable way, which makes this an error-prone string entry.

modules/apps/kate.nix
Asqiir: I changed to to a derivation in the same style. Imo it looks a bit bloated:

```nix
    checkThemeNameScript = pkgs.writeShellApplication {
      name = "checkThemeName";
      runtimeInputs = with pkgs; [ jq ];
      text = builtins.readFile ./check-theme-name-free.sh;
    };

    checkThemeName = name:
    ''
      ${checkThemeNameScript}/bin/checkThemeName ${name}
    '';

    script = pkgs.writeScript "kate-check" (checkThemeName cfg.editor.theme.name);
```

Maybe it would be nicer to do this with just 1-2 definitions? My attempt would be:

```nix
    checkThemeNameScript = pkgs.writeShellApplication {
      name = "checkThemeName";
      runtimeInputs = with pkgs; [ jq ];
      text = builtins.readFile ./check-theme-name-free.sh;
    };

    script = pkgs.writeScript "kate-check" "${checkThemeNameScript}/bin/checkThemeName ${name}";
```

What do you think?

modules/apps/kate.nix
Asqiir: If this means "make the lines shorter, introduce a linebreak": done

Asqiir:
@magnouvean do my changes look alright to you?

magnouvean:
I will check this out later today (when I get home). There has been a change in the configuration syntax lately after #94, so will probably need some tweaks, but they should be fairly straight forward :). Your solutions to the comments look good though :)

Asqiir:
@magnouvean You are right it needs some tweaks. I would have preferred to rebase this branch myself & do a new commit at the same time, so there wouldn't have been this broken state now ;)

magnouvean:
Ah sorry about that. Some prefer different things. I can revert it :)

modules/apps/default.nix
```diff
@@ -3,6 +3,6 @@
 {
   imports = [
     ./konsole.nix
-    ./kate.nix
+	./kate
```
magnouvean: Weird indenting here (run `nixpkgs-fmt` if you use this, or manually edit if you prefer)

modules/apps/kate/default.nix
```diff
@@ -0,0 +1,192 @@
...
+      '';
+      type = lib.types.str;
+      default = "";
+    };
+  };
+
+  config.programs.kate.editor.theme = {
+    name = lib.mkIf (cfg.enable && null != cfg.editor.theme.src) (lib.mkForce (builtins.fromJSON (builtins.readFile cfg.editor.theme.src))."metadata"."name");
+    # kate's naming scheme is ${themename}.theme
+    # which is why we use the same naming scheme here
```
magnouvean, line 166: Again I would prefer comments above

modules/apps/kate/default.nix
```diff
@@ -0,0 +1,192 @@
...
+
+  config = {
+    home.packages = lib.mkIf (cfg.enable && cfg.package != null) [ cfg.package ];
+
+    # In case of using a custom theme, check that there is no name collision
+    home.activation.checkKateTheme = lib.mkIf (cfg.enable && cfg.editor.theme.src != null) (lib.hm.dag.entryBefore [ "writeBoundary" ]
+    ''
+      $DRY_RUN_CMD ${script}
+    '');
+#       ${./check-theme-name-free.sh} ${cfg.editor.theme.name} ${pkgs.jq}/bin/jq
```
magnouvean: Remove this

magnouvean (commented):
Would be good if you could make sure to have ran `nixpkgs-fmt` on these files. Other than that if you can get these functional with the latest syntax I think this looks great!

modules/apps/kate/default.nix
magnouvean: I am talking about the comment only of course, not sure why github sometimes likes to add more lines :)

Asqiir: done

modules/apps/kate/default.nix
Asqiir: "above" as in "the comment line before the code line"?
Done

modules/apps/default.nix
Asqiir: I wasn't aware of `nixpkgs-fmt`, I ran it now on the file.

Asqiir:
I did the changes you proposed. 
The kate config now sets all values but makes none immutable. (Should it? What do we want the default behaviour to be?)

magnouvean:
No, immutability isn't needed granted the module works just fine without them. I see there were some problems with the merging leading to changes in `files` and `hotkeys` which shouldn't be there (but this is very easy to fix, I'll do it soon). Are you happy enough with this that we can merge then?

Asqiir:
@magnouvean yes I'm happy with the current state :)
thx for the fast workflow!

magnouvean:
Great, I'll merge right away then! Thanks for the contribution, prs like this helps `plasma-manager` grow and makes the plasma experience in nix better!

PR #99, opened 2024-03-29
fix incorrect `lib.mkIf` causing `autostart/plasma-manager-autostart.desktop` to fail

Kareem-Medhat:
Previous code was using an incorrect condition which always evaluated to true resulting in creating a startup service `autostart/plasma-manager-autostart.desktop` that would fail as they're were no script files to loop over in the `run_all.sh` returning a non-zero error code.


magnouvean:
Nice spot, thanks for fixing this. I'll merge right away

Issue #57, opened 2024-01-15
Example `flake.nix` has deprecated attributes `extraModules` and `configuration`


manuelbb-upb:
The example flake uses deprecated attributes `extraModules` and `configuration`, see https://nix-community.github.io/home-manager/release-notes.xhtml#sec-release-22.11-highlights
Putting both `plasma-manager` and `home.nix` in `modules` seems to work, though (at least it builds)...
Here is my flake:
```
{
  description = "Home-Manager Flake";

  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";

    home-manager.url = "github:nix-community/home-manager";
    home-manager.inputs.nixpkgs.follows = "nixpkgs";
    
    plasma-manager.url = "github:pjones/plasma-manager";
    plasma-manager.inputs.nixpkgs.follows = "nixpkgs";
    plasma-manager.inputs.home-manager.follows = "home-manager";

    nixgl.url = "github:guibou/nixGL";
    
    custom-any-nix-shell.url = "github:manuelbb-upb/any-nix-shell/nix_develop";
    custom-any-nix-shell.inputs.nixpkgs.follows = "nixpkgs"; 
  };

  outputs = inputs@{ nixpkgs, home-manager, plasma-manager, nixgl, custom-any-nix-shell, ... }:
  let
    system = "x86_64-linux";
  in {
    homeConfigurations.manuelbb = home-manager.lib.homeManagerConfiguration {
      pkgs = import nixpkgs {
        inherit system;
        config.allowUnfree = true;
        overlays = [ nixgl.overlay ];
      };
      modules = [ 
        inputs.plasma-manager.homeManagerModules.plasma-manager
        ./manuelbb_config/home.nix 
      ];
      extraSpecialArgs = { inherit inputs; };
    };
  };
}
```

Will test a bit more and hopefully manage a PR.

Asqiir:
I think this is fixed by now, at least I couldn't find any example still using `extraModules`.

Issue #84, opened 2024-03-11
How to use it?


igorcafe:
Sorry for the noobness, but could you provide an example of how to integrate it with home manager?

FF-AntiK:
You can find an example [here](https://github.com/pjones/plasma-manager/tree/trunk/example).

Asqiir:
@igoracmelo If you think anything specific in terms of an example is missing I can try to come up with one.

igorcafe:
@FF-AntiK @Asqiir how can I use it without flakes?

dvaerum:
I am new to Nix(OS) and home-manager, but I think the example code is outdated?

From what I can see here: https://nix-community.github.io/home-manager/release-notes.xhtml#sec-release-22.11-highlights

Something like `extraModules` does not exist in the newer versions of home-manager.

Xyz00777:
> @igoracmelo If you think anything specific in terms of an example is missing I can try to come up with one.

how can i implement the channel directly into my homeassistant/nix config?
i have done it with home assistant like these in my /etc/nixos/configuration.nix:

```
{ config, pkgs, lib, ... }:

let
  home-manager = builtins.fetchTarball "https://github.com/nix-community/home-manager/archive/master.tar.gz";
in
{
  imports =
    [ # Include the results of the hardware scan.
      (import "${home-manager}/nixos")
      ./hardware-configuration.nix
    ];

  # Home manager
  home-manager.users.xyz00777 = {
    imports = [
      ./homemanager/home.nix
    ];
  };

````

i also tried it the same way for plasma-manager but i didnt got it to work. In my opinion these is mutch nicer because when i have to setup my system again i dont have to think about adding the correct channel because these would do it more or less automaticly

igorcafe:
Guys, I haven't used the project and I don't plan to use it in the near future.
Feel free to close this whenever you are done :)

magnouvean:
Sorry for late response. The [getting started section in the README](https://github.com/pjones/plasma-manager/?tab=readme-ov-file#getting-started) offers some help in usage, mainly through examples.

Especially for home-manager setups without flakes I can recommend [the example directory for this](https://github.com/pjones/plasma-manager/tree/trunk/examples/homeManager) (it even includes some instructions), while for flake-based setups there are also a couple examples (see the getting started section).

I'll close this now, but if there are any problems related to these examples feel free to open a separate issue regarding this.

PR #100, opened 2024-03-30
fix example

Asqiir:
This is a fix for the errors noticed by @dvaerum in #84.

I had to update the stateversion (18.09 would have been default, but it did not compile with the old version but does with the new).

The example now works again.

### DIFF
--- a/example/flake.lock
+++ b/example/flake.lock
@@ -7,32 +7,32 @@
         ]
       },
       "locked": {
-        "lastModified": 1655880690,
-        "narHash": "sha256-pO+EqUefNFolIWV1DHXPfI8NuUtD4fd7JFKoaCjVXOs=",
+        "lastModified": 1710888565,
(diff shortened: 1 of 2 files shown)


Asqiir:
This should also close #57.

Asqiir:
@magnouvean I hope it's ok to ping you (or should I just wait with PRs like that?)

magnouvean:
Yeah no problem about it. I have been reworking the documentation and examples a bit in my fork at https://github.com/magnouvean/plasma-manager/tree/documentation-updates, and among the changes have been in flake.nix. There I've rewritten the flake.nix quite a lot. I think it should be ready in a few days and we do one big pr for improved documentation. Therefore I think there's not much point in fixing this right now as it probably will just be changed in a couple days anyway. I'll open a draft pr right away and we can work there directly

Asqiir:
Closing this in favour of #104

PR #104, opened 2024-03-30
Documentation updates

magnouvean:
The documentation needs some updates. This will pr will aim to vastly improve and update the documentation


Asqiir:
@magnouvean I left you a PR to your branch in your repo (is this the way to contribute to your draft PR?) bc I think we should provide both a "homemanager standalone" and a "homemanager as a module" example

README.md
```diff
@@ -99,20 +80,25 @@ the following shell command:
 nix run github:pjones/plasma-manager
 ```
 
-## Contributions and Maintenance
+## Getting started
```
bew: Have you considered moving the Getting Started section first?
Or maybe below the list of what we can do with plasma-manager?
In all cases it should be _before_ sections talking about the configuration itself (which needs some basic code infrastructure in your own code, that should be given by the Getting Started section)

magnouvean: I agree, I think that makes more sense

PR #107, opened 2024-03-31
docs: add installation instructions to README.md

nat-418:
This should make it easier for people to get started.

### DIFF
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -95,8 +95,7 @@ in
           cfg.workspace.colorScheme != null ||
           cfg.workspace.cursorTheme != null ||
           cfg.workspace.lookAndFeel != null ||
-          cfg.workspace.iconTheme != null ||
-          cfg.workspace.wallpaper != null))
+          cfg.workspace.iconTheme != null))
       {
         # We create a script which applies the different theme settings using
         # kde tools. We then run this using an autostart script, where this is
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,6 +99,38 @@ the following shell command:
 nix run github:pjones/plasma-manager
 ```
 
+## Getting Started
+
+Add this repository as a channel:
+
+```nix
+$ nix-channel --add https://github.com/pjone/plasma-manager/archive/trunk.tar.gz plasma-manager
```
TimekillerTK: Typo `pjone` -> `pjones`

TimekillerTK:
Just happened to be working on a new VM, so I tested the instructions out by copy/pasting, and they work - minus the typo.

You might want to add `home-manager switch` at the end to apply the config for completeness.

magnouvean:
I agree that having documentation using channels instead of flakes would be welcome (a lot of people don't use flakes and they should easily be able to use plasma-manager as well). I think this will need some changes though as the README was just changed yesterday. Also a question would be if plasma-manager was updated if `nix-channel --update` would actually update plasma-manager to the latest version using the above instructions.

nat-418:
> I agree that having documentation using channels instead of flakes would be welcome (a lot of people don't use flakes and they should easily be able to use plasma-manager as well). I think this will need some changes though as the README was just changed yesterday. Also a question would be if plasma-manager was updated if `nix-channel --update` would actually update plasma-manager to the latest version using the above instructions.

it just pulls whatever is in `trunk`

magnouvean:
The way this is right now there are two "getting started" sections in the README. I think it makes more sense to move the configuration into it's own file in the `examples` directory, for example to `examples/homeManagerNoFlake/home.nix` and then add the instructions to a README in that directory (for example in `examples/homeManagerNoFlake/README.md`). Then we can link this directory from the project README like the other examples. 

If this is done and the configuration you have provided works on your side I'll merge when this is done :)

nat-418:
> The way this is right now there are two "getting started" sections in the README. I think it makes more sense to move the configuration into it's own file in the `examples` directory, for example to `examples/homeManagerNoFlake/home.nix` and then add the instructions to a README in that directory (for example in `examples/homeManagerNoFlake/README.md`). Then we can link this directory from the project README like the other examples. 
> 
> 
> 
> If this is done and the configuration you have provided works on your side I'll merge when this is done :)

Shouldn't it be the opposite, as flakes are an experimental feature? Seems to me their use is a niche edge-case, not the default. This is at least how Home Manager itself treats the question: https://nix-community.github.io/home-manager/index.xhtml#ch-installation

magnouvean:
I'm not sure I get what you mean by opposite. I think it's better to present flakes and non-flakes the same way. Take a look at the [current README getting started section](https://github.com/pjones/plasma-manager?tab=readme-ov-file#getting-started) to see how the getting started section is structured. I think presenting the non-flake solution the same way makes  the most sense. That's what I meant by putting the text-section in `examples/homeManagerNoFlake/README.md` and the configuration in `examples/homeManagerNoFlake/home.nix` (and then adding a link to `examples/homeManagerNoFlake` from the getting started section).

Also my impression is that most people who use plasma-manager use the flakes version (that's what the original creator used if I remember correctly), but I think having tutorials/examples for both makes the most sense, and that they both are treated on the same level.

nat-418:
> I'm not sure I get what you mean by opposite.

I mean I think flakes should be put in the example configuration directory and not in the "Getting Started" section, as they are an experimental feature.

magnouvean:
Ah I see. I still think it's the best to keep all of them in the example directory and just link to them all. Flakes has sort of been the main way of using `plasma-manager` and hasn't really ever been viewed as an experimental feature for this project, but in principle I get what you mean. We could link the non-flake solution in the top of the getting started section though, which would sort of highlight this method above the "experimental" flake solutions.

nat-418:
I don't know why this was closed. I certainly did not do it intentionally. I guess I will make a new PR.

nat-418:
test

PR #118, opened 2024-04-07
Add separate wallpaper script with priority 3

magnouvean:
Fixes bug which reset the wallpaper when specifying screens of the panels. The problem was caused by the panel application script having to reset kwin, which also reset the wallpaper. This was fixed by setting the wallpaper after the panel application script has been ran.


noisesfromspace:
Thanks for your work on this @magnouvean

PR #121, opened 2024-04-10
add lsp server support to kate

Asqiir:
This PR adds LSP ("language server protocol") settings to kate.

It contains 2 preconfigured LSP servers – [nil]() for nix and [marksman]() for markdown. Users can add custom settings as well.

This PR is marked as a draft since I want to see whether these preconfigured LSP servers (nil and marksman) might be accepted into the kate project. If yes, we do not need these here.

### DIFF
--- a/modules/apps/kate/default.nix
+++ b/modules/apps/kate/default.nix
@@ -198,7 +198,9 @@ in
     default = {};
     type = lib.types.attrs;
     description = ''
-      Add more lsp server settings here. Check out the format on the [KDE page](https://docs.kde.org/stable5/en/kate/kate/kate-application-plugin-lspclient.html). Note that these are only the settings, the packages have to be installed separately.
+      Add more lsp server settings here. Check out the format on the
+      [KDE page](https://docs.kde.org/stable5/en/kate/kate/kate-application-plugin-lspclient.html).
+      Note that these are only the settings, the packages have to be installed separately.
     '';
   };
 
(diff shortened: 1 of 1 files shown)


Asqiir:
They indeed accepted it!
https://invent.kde.org/utilities/kate/-/merge_requests/1456

So it will probably be included in the next stable release of kate, 24.08 if I understand their release cycle correctly. Who knows when that one arrives in nixos-unstable, given that it is still on 23.08 despite 24.02 being already out for a while.

For this PR that means that I will undo most of the changes and keep just the extra settings.

modules/apps/kate/default.nix
```diff
@@ -189,4 +189,20 @@ in
...
+
+
+  # ==================================
+  #     LSP Servers
+
+  options.programs.kate.lsp.customServers = lib.mkOption {
+    default = {};
+    type = lib.types.attrs;
+    description = ''
+      Add more lsp server settings here. Check out the format on the [KDE page](https://docs.kde.org/stable5/en/kate/kate/kate-application-plugin-lspclient.html). Note that these are only the settings, the packages have to be installed separately.
```
magnouvean: Split this description into multiple lines (`gqq` this line if you are using vim)

magnouvean (commented):
Very cool that you managed to get this into kate. The pr looks good to apart from the one small comment. Do you reckon this is ready to be merged right away after this is fixed?

Asqiir:
Yes, I think it is ready now

PR #122, opened 2024-04-10
docs: add standard Nix / Home Manager example

nat-418:


magnouvean:
Well solved. Thanks for this, I think it will make the project more approachable :)

Issue #47, opened 2024-01-09
Configuring LibInput devices


billy4479:
I happen to use natural scrolling and a few other options on my laptop touchpad.
These option are set in `~/.config/kcminputrc` as 

```
[Libinput][10182][480][GXTP7863:00 27C6:01E0 Touchpad]
DisableWhileTyping=false
NaturalScroll=true
TapToClick=true
```

I'd like to use these same settings for any laptop I (will) use.
As you can see the group name is kinda device-specific: for sure the last part (`[GXTP7863:00 27C6:01E0 Touchpad]`) is the name of the input device as it shows in `libinput list-devices`; I still haven't figured out what the other two numbers are (`[10182]` and `[480]`). 

For now I can just hardcode them in my config but I don't think that's the best solution..

If someone manages to figure out what the other two numbers are we could query them at activation-time and have an option to enable or disable such features for any touchpad.

toast003:
Tried finding out what writes the numbers are by reading the [source code for the mouse kcm](https://invent.kde.org/plasma/plasma-desktop/-/tree/master/kcms/mouse) but I haven't gotten anything yet

toast003:
[This repo](https://github.com/MatMoul/plasma-config-matmoul/blob/master/profile/.config/kcminputrc) just puts setting under `[Mouse]` :thinking: 
I'll try it out 
EDIT: Don't think it works anymore (the config file I linked is 8 years old after all), but I'm doing other things so I can't log out or restart plasma, so maybe that's why it doesn't work

billy4479:
> Tried finding out what writes the numbers are by reading the [source code for the mouse kcm](https://invent.kde.org/plasma/plasma-desktop/-/tree/master/kcms/mouse) but I haven't gotten anything yet

Thanks for the link, I'll try to look into it too.

Also, I forgot to add in the original post that I'm using the Wayland session. Maybe the X11 session uses some other config (?), I'll try to login into a X session when I have time and I'll let you know if I find any difference.

toast003:
I'm also using Wayland 

11 Jan 2024, 12:09 pm by ***@***.***:

ribru17:
> If someone manages to figure out what the other two numbers are we could query them at activation-time and have an option to enable or disable such features for any touchpad.

The first two numbers are the decimal representations of the middle (hexadecimal) numbers: e.g., for
`[Libinput][10182][480][GXTP7863:00 27C6:01E0 Touchpad]`:
[...]

PR #123, opened 2024-04-11
Add new input module

toast003:


toast003:
Had a bit of a brain fart there, sorry for the force pushing

toast003:
@magnouvean I changed the write_config script a bit, since doing base conversion inside nix seems too complicated.
I am very new at python, and although my code seems to work just fine I would appreciate if you could take a look at it

toast003:
Also I don't think the check failing was due to something I did, not sure why
EDIT: nvm it is, I'll fix it later

script/write_config.py
```diff
@@ -76,6 +77,21 @@ def escape(s: str) -> str:
             s[i] = "\\s"
     return "".join(s)
 
+# Some values need to be "transformed", such as being converted from hex to decimal
+def transformValues(dOrig):
+    d = dOrig
```
magnouvean: Not sure if this is needed, since this won't copy anything, it will just be a pointer to `dOrig`. You can use `dOrig.copy()` if this is wanted (which I think makes sense in this case).

script/write_config.py
```diff
@@ -76,6 +77,21 @@ def escape(s: str) -> str:
...
+    cPath = os.path.expanduser("~/.config/kcminputrc")
+    if cPath in d:
+        for itemStr in list(d[cPath]):
+            if itemStr.startswith("Libinput/"):
+                item = d[cPath].pop(itemStr)
+                itemStrList = itemStr.split("/")
+                itemStrList[1] = str(int(itemStrList[1],16))
+                itemStrList[2] = str(int(itemStrList[2],16))
+                itemStr = "/".join(itemStrList)
+                d[cPath][itemStr] = item
```
magnouvean: In python it is standard practice to use snake_case, not camelCase (unlike nix, java, C# etc). I think this should be the case here as well.

script/write_config.py
```diff
@@ -329,7 +345,8 @@ def main():
 
     reset_files = set(sys.argv[2].split(" ")) if sys.argv[2] != "" else set()
     immutable_by_default = bool(sys.argv[3])
-    d = json.loads(json_str)
+    dRaw = json.loads(json_str)
+    d = transformValues(dRaw)
```
magnouvean: snake_case is better here as well

modules/input.nix
```diff
@@ -0,0 +1,104 @@
...
+    };
+    productId = mkOption {
+      type = with types; nullOr str;
+      default = null;
+      example = "21128";
+      description = ''
+        The product ID of the trackpad
+        You can find it out running cat /proc/bus/input/devices | grep -i touchpad
+      '';
+    };
```
magnouvean: These three have allowed values as `null` but if the touchpad is enabled and these are not set I'm pretty sure it will crash. This should probably be checked in an assertion so we could give out a proper error message.

modules/input.nix
```diff
@@ -0,0 +1,104 @@
...
+            LayoutList.value = strings.concatStringsSep "," cfg.input.keyboard.layouts;
+          };
+        }
+      )
+    ]
+  );
+
+  # Trackpad options
+
+  options.programs.plasma.input.touchpad = {
```
magnouvean: The options here will write to the same group as mentioned in #47, am I right? If so we should probably add `DisableWhileTyping`, `NaturalScroll` and `TapToClick` as options which would close that issue.

magnouvean (commented):
I can't really test much atm, but looking at the code in general this looks fine after the minor suggested tweaks :)

magnouvean:
Also https://github.com/nix-community/plasma-manager/blob/d12cb0ed42a81ff84330b5cc94bff6abfeeb6da8/README.md?plain=1#L53 should probably be edited or removed when we're done with this pr.

script/write_config.py
toast003: This is my first real python code and I also happen to come from a java/c# background so I didn't know this, will change now

toast003:
I will when I finish

modules/input.nix
toast003: Yes!
I was going to add those first, but I thought that making the user look for the IDs of the trackpad, and then turning them to decimal was a bit too dumb. 

Already took care of the hexadecimal to decimal part here, but I'll do the "looking for the IDs" part in a separate pr, since thinking about how to do that is what burnt me out, and I don't want this pr to be stalled for another month :(

magnouvean: No worries, I think just getting this into `plasma-manager` gets us a long way.
Looking for the IDs isn't necessarily going to be that easy, and would, if done at activation-time not be reproducible either (since the output would depend on the devices attached when the activation is applied). I think when doing that we could maybe consider just altering the config-file generated automatically by kde, which contains these ids aleady, in a startup-script, as that would be declarative, and maybe easier to do. I would maybe require a reload of some kde component, but `plasma-manager` should support that already.

Anyway that will be in the future, there's no need to think about this yet (and I think that it isn't necessarily a must for this module anyway).

magnouvean:
I'll try to add support for configuring multiple touchpads and add some assertions today. I think I'll just commit directly here if that's OK :)

magnouvean:
Imo it's more or less good enough to merge in the current state. I haven't found any issues (though I also haven't done enough testing to confirm a lot of edge-cases). There are obviously some tweaks as mentioned above such as changing the README and maybe also adding some examples.

HeitorAugustoLN:
@toast003 I sent you a pull request that adds configurations for mouses in the input module.

toast003:
You guys did all of my work :sweat_smile: 

@HeitorAugustoLN I'll take a look at it rn

toast003:
All seems good to me!

I'll change the readme and after that I think we can merge the PR

modules/input.nix
```diff
@@ -0,0 +1,386 @@
...
+          Enables or disables the mouse
+        '';
+      };
+      name = mkOption {
+        type = types.str;
+        default = null;
+        example = "Logitech G403 HERO Gaming Mouse";
+        description = ''
+          The name of the mouse.
+          You can find it out running cat /proc/bus/input/devices | grep -B 1 -i mouse
```
magnouvean: This doesn't work for me because my mouse doesn't have anything with mouse in the name. `grep '^H:.*mouse' -B 5 /proc/bus/input/devices` seems to work for me though. Seemingly this could maybe also be considered for the touchpad as well since I'm not sure all touchpads include `touchpad` in the name, but they should have `mouse*` in the handlers section

magnouvean (commented):
I'll test more right away

modules/input.nix
HeitorAugustoLN: image
It works for me. But it is probably not a general rule.

magnouvean: Yeah that works because the mouse has "mouse" in the name, but for mine (a logitech mouse as well) it doesn't have this. In your case though the command I provided should work as well though?

modules/input.nix
```diff
@@ -0,0 +1,386 @@
...
+    let
+      mouseName = mouse.name;
+      mouseVendor = mouse.vendorId;
+      mouseProduct = mouse.productId;
+    in
+    {
+      "Libinput/${mouseVendor}/${mouseProduct}/${mouseName}" = {
+        Enabled = mouse.enable;
+        LeftHanded = mouse.leftHanded;
+        MiddleMouseEmulation = mouse.middleMouseEmulation;
```
magnouvean, line 282: I also think it should be `MiddleButtonEmulation` here

magnouvean:
But yeah other than the two minor things above this work great so we can probably merge after they are fixed :)

modules/input.nix
HeitorAugustoLN: Your command works too. 
image

modules/input.nix
HeitorAugustoLN: You're right, I mixed up the names

HeitorAugustoLN: @toast003 could you fix this?

modules/input.nix
magnouvean: Great. My suggestion to solve this would be to replace the old commands (in the three descriptions for both mouse and touchpad) with `grep '^H:.*mouse' -B 5 /proc/bus/input/devices`. We might be able to get away with the old command for touchpads, but afaik it's no requirement for those to have `touchpad` in the name either so I think the new command is still better for touchpads as well.

toast003: For trackpads you can just grep  `/proc/bus/input/devices` since most trackpads are i2c devices and (afaik) i2c devices don't have names, so trackpads always have trackpad in the name.
For mice we could just check `lsusb`
image

modules/input.nix
toast003: Done

HeitorAugustoLN: Thanks!

modules/input.nix
magnouvean: On my end when running the grep-command as it is now it matches `ETPS/2 Elantech Touchpad` which is in the name section of the touchpad in `/proc/bus/input/devices`, and if it was called something else it may not have matched. With that said I'm not sure if it's a real issue as most touchpads probably contain "touchpad" in some form in the name (unless it's called something with trackpad or something).
As for `lsusb` there I can see two potential problems, the first being that `lsusb` isn't installed on most nixos machines I don't think, so it's less accessible than just using `/proc/bus/input/devices`, and second it may not support bluetooth mice (I don't know if it does). The command provided above does just uses `/proc/bus/input/devices` and would work for both touchpads and mice

toast003: While I don't think your first problem is that big of a deal (you can use comma or `nix run`), I do agree with the second.
lsusb only lists usb devices, so it won't catch anything else. For now I would say to leave both, but we could consider making a python script that figures it out for you

magnouvean: Yeah figuring this out automatically could be done through some startup script or something and is a task for later. My point was that the descriptions for the options should be correct/helpful. For that the grep command above would suffice for now I reckon, or alternatively we could just add "see /proc/bus/input/devices for the values" or something. A.t.m. the grep commands in the descriptions for the mice doesn't work and the grep commands for the touchpad only works for some hardware (only when the name contains touchpad), and I think this should be correct before we merge into `trunk`.

HeitorAugustoLN: I think a more general description like `see /proc/bus/input/devices for the values` would be better

script/write_config.py
```diff
@@ -76,6 +77,21 @@ def escape(s: str) -> str:
...
+    if c_path in d:
+        for item_str in list(d[c_path]):
+            if item_str.startswith("Libinput/"):
+                item = d[c_path].pop(item_str)
+                item_str_list = item_str.split("/")
+                item_str_list[1] = str(int(item_str_list[1],16))
+                item_str_list[2] = str(int(item_str_list[2],16))
+                item_str = "/".join(item_str_list)
+                d[c_path][item_str] = item
+    return d;
```
apfelkuchen6, line 94: I don't really like this because it changes the interface of the `configFile` option in an unexpected manner. It's fine if we want to have the options of the input module expect the ids in hex, but then the conversion should also be done in the module (possibly with something like [this](https://github.com/Misterio77/nix-colors/blob/b92df8f5eb1fa20d8e09810c03c9dc0d94ef2820/lib/core/conversions.nix#L72-L94)).

This change broke my config, which included `programs.plasma.configFile.kcminputrc."Libinput/1390/269/ELECOM TrackBall Mouse HUGE TrackBall".ScrollMethod.value =  4;`. Now the ids are silently mangled.

magnouvean: I agree there. It should be fairly straight forward to do so as well. I'll try to do the hex-conversion in nix instead and open a PR about it.

HeitorAugustoLN: >  but then the conversion should also be done in the module

This would be indeed better.

magnouvean: Should be fixed now after #295.

Issue #119, opened 2024-04-07
Volume keys not working after switching config


valeth:
Some default shortcuts no longer function after switching to a new home manager configuration.
They start working again after the `plasma-plasmashell` systemd user service is restarted.

This only seems to happen after a switch, just rebooting into the system normally works as intended.

I started noticing this problem in Plasma 6, but I also haven't used PM with Plasma 5 long enough to confirm that only that version is affected.

magnouvean:
I also have seen this. Do you use `overrideConfig` by any chance? If so I would be inclined to believe it could have something to do with the deletion of some rc-files in `~/.config`. It wasn't a problem in plasma 5 as far as I could tell.

valeth:
Yes, I am using `overrideConfig`

magnouvean:
Yeah I'm then quite confident it's related to that. A fix shouldn't be that hard though as it should just be to write the necessary lines which handles this functionality to some file instead of deleting the file completely as we do a.t.m.. Finding out which options causes this is a little worse though, but should be doable with a little digging.

DarkKronicle:
I'm having the same issue, although mine won't work after I boot. I think I have an idea what causes it. Volume keys require a volume control applet to exist (this can also be contained within the systemtray). When plasma manager reconstructs the panels, it removes the old ones and creates new ones on top of it. I think this process results in the original volume applet having control over the keybinds, even though it no longer exists.

This is a known issue and it seems like it will be fixed soon (probably 6.1 release).

- [KDE Bug](https://bugs.kde.org/show_bug.cgi?id=391578)
- [Relevant Merge](https://invent.kde.org/plasma/plasma-pa/-/merge_requests/221) (3 weeks ago)

Temporarily this can be fixed by running:
```js
var panel = panels()[0];
panel.addWidget("org.kde.plasma.volume");
```
or just adding the widget manually. I'll see if adding this to a script or something fixes this.

DarkKronicle:
Here's an incredibly jank fix that I've gotten to work: On boot overwrite `plasma-org.kde.plasma.desktop-appletsrc` with a version that doesn't have a systemtray or volume control applet. Then when plasma manager creates the panels, there will be no issue with previous volume controls not existing.

valeth:
Noticed another small glitch that is again fixed by restarting the plasmashell service.

Right after switching config and restarting (same as with the volume control issue) the system tray is "zero-sized", meaning it only shows up as a dot, probably because the widgets have borders.
[...]

PR #127, opened 2024-04-13
Add support for persistent keys

magnouvean:
Should address the persistence we have discussed in #120. To allow for the functionality I had to do quite a bit of refactoring, and therefore this is quite a big change moving some of the nix code to the python script. Therefore there could also be some bugs introduced (though none I have found just yet). I very much welcome anyone to test on their own configs and let me know how it works before we merge (it's probably wise to backup the config-files just in case some bug I have overseen). This also hopefully should open up for solving #119 more easily as we can just enable persistence for some of the keys (granted we can find out which ones are at fault).

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -27,6 +27,14 @@ let
           adding [$e] to the end of the key.
         '';
       };
+      persistent = lib.mkOption {
+        type = bool;
+        default = false;
+        description = ''
+          When overrideConfig is enabled and the key is persistent,
+          plasma-manager will leave it unchanged after activation.
+        '';
+      };
     };
   });
 in
--- a/lib/writeconfig.nix
+++ b/lib/writeconfig.nix
@@ -1,4 +1,4 @@
-{ pkgs, lib }:
+{ pkgs, lib, config }:
 
 let
   writeConfigScript = pkgs.writeShellApplication {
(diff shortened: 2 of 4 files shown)


lib/types.nix
```diff
@@ -27,6 +27,14 @@ let
           adding [$e] to the end of the key.
         '';
       };
+      persistent = lib.mkOption {
```
Asqiir, line 30: Does this allow the user to set `persistent = true;` and `immutable = true;` at the same time? What is the desired behaviour of plasma-manager in that case?
(Imo this shouldn't be possible.)

Asqiir, line 30: And what about 
```nix
somekey = {
  value = "myval";
  persistent = "true";
};
```
Which value should be set on activation?

magnouvean: I agree, it shouldn't be possible. It's an easy fix luckily

magnouvean: As of the latest version I am pretty sure this isn't possible to do. It won't allow values when persistence is enabled as it doesn't really make sense.

Asqiir: that's great to hear :)

question:
doesn't that mean there are only 4 options left(?):
1. no value -> persistent by default
2. immutable and a value
3. immutable without a value (should that be possible?)
4. persistent and no value, explicitly

if option 3 should not be possible we could do something like: (just a suggestion, maybe for later)

```nix
somekey = null; # means persistent, is the default
someotherkey = "value"; # makes it immutable
```

then we do not need the immutable/persistent/value flags anymore (?)

magnouvean: 3 shouldn't have been possible, but there wasn't any checks in the code which made it illegal (but just before you commented I added a check for this as well :)). There also is shell-expansion of variables, which pretty much is the same as immutability. I also think it is possible to the explicitly setting the key to `null` which should remove the key from the file altogether. There's plenty of options and combinations, but I think we have gotten rid of most of the combinations that don't make sense.

magnouvean:
Did the functionality work the way you expected @Asqiir? In the current state it works for my needs anyway, so if there isn't anything I have missed I think we can merge this quite soon.

Asqiir:
It does!
And I see now that leaving room for more flags makes sense :)
Thanks for building this

PR #141, opened 2024-04-22
Fix typo in example home.nix

abhr-0:


magnouvean:
Thanks, will merge right away.

PR #152, opened 2024-04-28
Add support for wallpaper type Picture of the Day

abhr-0:


modules/workspace.nix
```diff
@@ -99,14 +108,27 @@ in
         Allows you to set wallpaper slideshow. Needs a directory of your wallpapers and an interval length.
       '';
     };
+
+    wallpaperPictureOfTheDay = lib.mkOption {
+      type = lib.types.nullOr wallpaperPictureOfTheDayType;
+      default = null;
+      example = "apod";
+      description = ''
+        Allows you to set wallpaper using the picture of the day plugin. Needs the provider and the fill mode.
```
magnouvean: I don't see anything to do with the fill-mode in this pr. Is something missing?

magnouvean (commented):
Looks good otherwise. Will test soon.

magnouvean:
Just tested and it does work well! Could just set the type to `nullOr (enum ["apod" "bing" "flickr" "natgeo" "noaa" "wcpotd" "epod" "simonstalenhag" ])` I reckon. Works fine as is now without fill-mode functionality.

modules/workspace.nix
```diff
@@ -99,14 +108,27 @@ in
...
+
   };
 
   config = (lib.mkIf cfg.enable (lib.mkMerge [
     {
       assertions = [
         {
-          assertion = (cfg.workspace.wallpaperSlideShow == null || cfg.workspace.wallpaper == null);
-          message = "Cannot set both wallpaper and wallpaperSlideShow at the same time.";
+          assertion = ((cfg.workspace.wallpaperSlideShow == null && cfg.workspace.wallpaper == null && cfg.workspace.wallpaperPictureOfTheDay != null)
```
pluiedev: Consider doing something like this instead: (see https://github.com/nix-community/home-manager/blob/master/modules/programs/git.nix#L360-L365)
```nix
assertion = let 
  wallpapers = with cfg.workspace; [wallPaperSlideShow wallpaper wallpaperPictureOfTheDay];
in
  lib.count (x: x != null) wallpapers <= 1;
```

Additionally, this assertion currently triggers when none of the three are set — are we really going to force people to pick a wallpaper?

modules/workspace.nix
```diff
@@ -99,14 +108,27 @@ in
         Allows you to set wallpaper slideshow. Needs a directory of your wallpapers and an interval length.
       '';
     };
+
+    wallpaperPictureOfTheDay = lib.mkOption {
+      type = lib.types.nullOr wallpaperPictureOfTheDayType;
+      default = null;
+      example = "apod";
```
pluiedev: Wait... I don't think you can assign strings to submodules. Honestly, the submodule isn't even necessary and a simple enum type would do just fine

modules/workspace.nix
```diff
@@ -170,5 +192,19 @@ in
         priority = 3;
       };
     })
+    (lib.mkIf (cfg.workspace.wallpaperPictureOfTheDay != null) {
+      programs.plasma.startup.desktopScript."set_wallpaper_potd" = {
+        text = ''
+          let allDesktops = desktops();
+          for (var desktopIndex = 0; desktopIndex < allDesktops.length; desktopIndex++) {
+              var desktop = allDesktops[desktopIndex];
```
pluiedev: It's currently the year 2024. Plasma's desktop scripts support ES6, even though all their docs are ES5 for some reason
```js
for (const desktop of desktops()) {
    // ...
}
```

pluiedev (changes requested):

modules/workspace.nix
abhr-0: Oh sorry, actually I had added support for fill-mode but specifically just for potd, but I realised that we should be able to set it for all types of wallpapers. So I removed that bit (cuz that should be a different PR) but ig I forgot to update the description. Whoops.

modules/workspace.nix
abhr-0: > ```nix
> assertion = let 
>   wallpapers = with cfg.workspace; [wallPaperSlideShow wallpaper wallpaperPictureOfTheDay];
> in
>   lib.count (x: x != null) wallpapers <= 1;
> ```

This! I really hated the way I did it, but I just couldn't figure out how to make it nicer.

> Additionally, this assertion currently triggers when none of the three are set — are we really going to force people to pick a wallpaper?

Oops. that's definitely wasn't what I wanted to do.

Big thanks for the help

modules/workspace.nix
abhr-0: Ah, yes that's completely true, I'll update the example.

abhr-0:
Big thanks to @magnouvean and @pluiedev for the help.

Honestly, I don't have much experience with nix (or even git) . I apologize for all the goof ups.

modules/workspace.nix
```diff
@@ -99,14 +108,27 @@ in
         Allows you to set wallpaper slideshow. Needs a directory of your wallpapers and an interval length.
       '';
     };
+
+    wallpaperPictureOfTheDay = lib.mkOption {
+      type = lib.types.nullOr wallpaperPictureOfTheDayType;
+      default = null;
+      example = ''{ provider = "apod" }'';
```
magnouvean: Should be `{ provider = "apod"; }` instead (i.e. remove the quotes and add semicolon after provider).

magnouvean (commented):
No worries, it's a good way to learn (and this helps get some new features for the project that I think I might even start to use myself).

Anyways this looks good to me now once the adjustment in this comment is fixed. Is it a possibility that other options (other than provider) is added to `wallpaperPictureOfTheDay`? If not we could just set the type of `wallpaperPictureOfTheDay` to `nullOr (enum ["apod" "bing" "flickr" "natgeo" "noaa" "wcpotd" "epod" "simonstalenhag" ]);` directly, and delete `wallpaperPictureOfTheDayType` altogether. I'm not opposed to having this here though if it makes sense to extend one day, as the current type does make sense.

modules/workspace.nix
abhr-0: Well, POTD seems to have another option to enable/disable updating over a metered connection. Adding it immediately.

> Should be `{ provider = "apod"; }` instead (i.e. remove the quotes and add semicolon after provider).

Even after being told that the example was not supposed to be a string, I fixed it...by changing it to another string. :sweat_smile:

magnouvean:
Very nice! I will test and hopefully merge soon if everything works fine

modules/workspace.nix
```diff
@@ -170,5 +197,20 @@ in
         priority = 3;
       };
     })
+    (lib.mkIf (cfg.workspace.wallpaperPictureOfTheDay != null) {
+      programs.plasma.startup.desktopScript."set_wallpaper_potd" = {
+        text = ''
+          let allDesktops = desktops();
+          for (const desktop of allDesktops) {
+              var desktop = allDesktops[desktopIndex];
```
pluiedev: This doesn't compile as you redefine desktop, and also, desktopIndex is undefined

modules/workspace.nix
```diff
@@ -170,5 +197,20 @@ in
...
+              desktop.wallpaperPlugin = "org.kde.potd";
+              desktop.currentConfigGroup = Array("Wallpaper", "org.kde.potd", "General");
```
pluiedev: Use array literal syntax
```js
["Wallpaper", "org.kde.potd", "General"]
```
I know that KDE docs use the Array(...) form but it's really just outdated. Writing modern JS works just fine.

pluiedev (changes requested):

pluiedev (approved):

magnouvean:
Just tested and works as expected still. Good work. LGTM

PR #153, opened 2024-04-28
module(panel): refactors and small fixes

pluiedev:
This PR refactors the logic used in generating the desktop script for panels to make it more readable, and adds a `lengthMode` setting that allows Plasma 6 users to configure their panels' length mode.

Additionally, there was a small bug in the desktop script generation logic — if you have two consecutive single quotes in the script, the escaped sequence will produce an invalid Bash script, for whatever reason. I've changed it to use a heredoc instead, which should be much more resilient.

### DIFF
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -21,7 +21,7 @@ let
     };
   };
 
-  panelType = lib.types.submodule {
+  panelType = lib.types.submodule ({config, ...}: {
     options = {
       height = lib.mkOption {
         type = lib.types.int;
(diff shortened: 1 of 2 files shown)


magnouvean:
It appears that some command returns a non-zero exit-code in the `2_desktop_script_apply_panels.sh` script leading the `success` variable to be set to `0` and no `last_run_desktop_script_apply_panels` file will be created (even though the script seems to be working just fine). I'll see if this is caused by something here or if this is a general problem.

magnouvean:
It seems to be the read command which is the problem. I'll see if we can create a layout.js file in the data-directory and just cat that file instead.

magnouvean:
That's done the trick :). Will merge now, thanks for this.

PR #154, opened 2024-04-28
modules(fonts): init

pluiedev:
This PR adds a new module for setting system fonts. I honestly spent way too much time reading how `QFont` works, but this module should at least be _very_ comprehensive to anyone who needs it.

Tested locally in my own system flake — works great!

I'm not exactly happy with how I handled the `pointSize` vs `pixelSize` thing, but submodules unfortunately can't have assertions... :(

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ At the moment `plasma-manager` supports configuring the following:
 - Shortcuts (via the `shortcuts` module)
 - Hotkeys (via the `hotkeys` module)
 - Panels (via the `panels` module)
+- Fonts (via the `fonts` module)
 - KDE apps (via the `apps` module). In particular the following kde apps have
   modules in `plasma-manager`:
   - kate
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -22,6 +22,13 @@
       command = "konsole";
     };
 
+    fonts = {
+      general = {
+        family = "JetBrains Mono";
+        pointSize = 12;
+      };
+    };
+
     panels = [
       # Windows-like panel at the bottom
       {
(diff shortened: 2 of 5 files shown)


magnouvean:
Wow nice work, looks good from what I can tell (although I won't even pretend to understand what half those values mean). We should add a note in the README that font-configuration is supported and maybe represent this module a bit in the examples as well. I'll test some of this as well, but may take a few days (though if you have tested this and it works the amount of testing probably doesn't need to be that much).

pluiedev:
> Wow nice work, looks good from what I can tell (although I won't even pretend to understand what half those values mean). We should add a note in the README that font-configuration is supported and maybe represent this module a bit in the examples as well. I'll test some of this as well, but may take a few days (though if you have tested this and it works the amount of testing probably doesn't need to be that much).

Yeah, realistically *most* people won't EVER have to touch any of these values — my own config only uses the font family and point size, which are the two only required ones anyway. Testing this is actually super easy with the `fontToString` function in place, and maybe we can just move it to a library module along with the ginormous amounts of internal tables and such and refer to it elsewhere xD

magnouvean:
Just tested and everything works perfectly on my end as well. I pushed a very simple example config in the example `home.nix` as it was missing. I see the check fails with this but this isn't the fault of this PR and will be fixed once we make some adjustments to the CI. Are you happy enough with the state of this PR that we can merge now @pluiedev?

pluiedev:
> Just tested and everything works perfectly on my end as well. I pushed a very simple example config in the example `home.nix` as it was missing. I see the check fails with this but this isn't the fault of this PR and will be fixed once we make some adjustments to the CI. Are you happy enough with the state of this PR that we can merge now @pluiedev?

I think so yeah :shipit:

PR #156, opened 2024-04-29
Add tests for `write_config.py` and fix discovered bugs

quentinmit:
(Note that this PR is on top of #144. When that PR is merged I'll rebase this onto `trunk`.)

This PR adds a number of tests in `test/basic.nix` and fixes many incorrect cases in `write_config.py`:

- Use KDE-specific escaping rules (including correct handling of leading and trailing spaces)
- Correctly handle keys that have changed from immutable to not or vice versa
- Correctly preserve existing keys that have escape characters
- Test setting keys in nested groups
- Test group names that need escaping
- Pre-existing top level keys

I restructured `write_config.py` to add a class called `ConfigValue`  that contains all of the conversions between KDE, Nix, and in-memory representations of config. I also changed it to use tuples to represent group names, so there are no magic strings in the in-memory data.

There's one case that I didn't touch in this PR, which is the case of setting top-level keys (not inside any group). That would require changing the user-visible API. `kreadconfig5`/`kwriteconfig5` use the magic string `<default>` to represent the "group" of top-level keys, if we want to support that same syntax. Or we could use an empty string (although I think that's also a valid group name?). At least now it should correctly round-trip keys set at the top-level outside plasma-manager.


quentinmit:
Pinging @magnouvean. This is the next followup to #144.

magnouvean:
I'll take a look. There seems to be quite a bit of new code to take a look at so it might take a little time to go through fully.

magnouvean:
Looks quite good. The only thing I noticed is that since we are now escaping keys that the higher-level interfaces, like shortcuts, won't support having `[$i]` or similar in the keys anymore. I'll see if I can add an option to support for example `{ key = "key"; immutable = true; }` for these modules as well.

magnouvean:
Actually, when thinking about it it's probably not even needed. Adding `[$i]` or similar never actually does anything for the files they are dealing with anyway.

magnouvean:
All works as expected, I'll merge right away. Good work!

quentinmit:
> Looks quite good. The only thing I noticed is that since we are now escaping keys that the higher-level interfaces, like shortcuts, won't support having `[$i]` or similar in the keys anymore. I'll see if I can add an option to support for example `{ key = "key"; immutable = true; }` for these modules as well.

Yeah, it was intentional that a higher-level `key[$i]` should get escaped to `key\x5b$i\x5d`. If you want an immutable value, we provide that in the form of `immutable = true`. I think a user could put `configFile.file.group.key.immutable = true' in their config alongside the higher-level module, though I agree that's not great since they'd need to manually calculate the keys that will be set.

Maybe some of the higher-level modules could be changed to use `coercedSettingsType`, so that the user could specify `{ value = 1; immutable = true; }` directly in the high-level module?

magnouvean:
We should probably consider it. Ideally when using higher-level modules the user shouldn't have to think about immutability at all. I'll see if any issue pops up requesting it first, as to my experience it's not really been needed for any of the higher-level modules. Another option could be to just set immutability by default in some of the higer-level modules if it helps. That way the user can also just expect it to work without thinking of immutability and such. We'll see.

PR #161, opened 2024-05-01
modules(panels): add widget-specific options

### LINKED ISSUE #155: Panels: Widget-specific settings
After learning how to configure the widgets for my panel, I feel like it would be better if we can have a more idiomatic layer of sorts that abstracts a lot of the ugliness away. For example, for system monitor, right now you need to write this, which is very tedious:
```nix
{
  name = "org.kde.plasma.systemmonitor";
  config = {
    Appearance.title = "CPU Usage";
    SensorColors."cpu/all/usage" = "250,179,135";
    Sensors = {
	  # The escaping is needed so that it fits into JS as a regular string
      highPrioritySensorIds = "[\\\"cpu/all/usage\\\"]";
      lowPrioritySensorIds = "[\\\"cpu/all/averageTemperature\\\", \\\"cpu/all/averageFrequency\\\"]";
      totalSensors = "[\\\"cpu/all/usage\\\"]";
	};
  };
}
```
What if we can write it more like:
```nix
{
  systemMonitor = {
    title = "CPU Usage";
    sensorColors."cpu/all/usage" = "250,179,135";
    totalSensors = ["cpu/all/usage"];
    sensors = ["cpu/all/usage"];
    textOnlySensors = ["cpu/all/averageTemperature" "cpu/all/averageFrequency"];
  };
}
```
Which is *much* easier to understand, IMO.
There's even already a mechanism in Nix's option system that does this: [`types.attrTag`](https://nixos.org/manual/nixos/unstable/#sec-option-types-sums)

pluiedev:
Adds widget-specific settings to widgets in panels.

Instead of writing this:

```nix
{
  name = "org.kde.plasma.digitalclock";
  config.Appearance = {
    dateFormat = "custom";
    customDateFormat = "d MMM ''yy / ddd";
    selectedTimeZones = ["Europe/Berlin" "Asia/Shanghai"];
    wheelChangesTimezone = "true";
    displayTimezoneFormat = "FullText";
    use24hFormat = "2";
  };
}
```

You can now write something more akin to this:

```nix
{
  digitalClock = {
    date.format.custom = "d MMM ''yy / ddd";
    time.format = "24h";
    timeZone = {
      selected = ["Europe/Berlin" "Asia/Shanghai"];
      changeOnScroll = true;
      format = "city";
    };
  };
}
```

Which is much easier to write, much easier to understand, much more type-safe (not everything is a string or a string list now), and doesn't require users digging deep into configuration files or KDE's scripting documentation.

The layout has been arranged so that it's very simple to add new widget configs — one just adds a new file under `modules/widgets/`, and add that to the sources array in `modules/widgets/default.nix`.

Each widget config is expected to be in the following structure:

```nix
{
  # This widget name has to be unique, as it's the only way users
  # can specify the type of the widget they're currently adding and configuring
  widgetName = {
    description = "A sample widget.";

    opts = {
      # List all the options the widget should have here.
      enableFoobar = mkEnableOption "foobar";
    };

    # Takes the widget settings as an input, and outputs an attrset with
    # name and config fields — just like a manually defined "raw" widget.
    convert = { enableFoobar }: {
      name = "org.kde.plasma.sample";
      config = {
        # Do note that this is not a module config.
        # You *cannot* use mkIf, mkMerge, etc here!
        Appearance.foobar = enableFoobar;
      };
    };
  };
}
```

And as long as this structure is followed, new widgets should just work in this system automatically.

There are two widgets attached to this PR as of now: the digital clock and system monitor. However, due to the aforementioned structure, it should be very easy to create subsequent PRs that add other widget types.

Fixes #155

### DIFF
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -61,7 +61,19 @@
           # widget will add them with the default configuration.
           "org.kde.plasma.marginsseparator"
           "org.kde.plasma.systemtray"
-          "org.kde.plasma.digitalclock"
+          # If you need configuration for your widget, instead of specifying the
+          # the keys and values directly using the config attribute as shown
+          # above, plasma-manager also provides some higher-level interfaces for
+          # configuring the widgets. See modules/widgets for supported widgets
+          # and options for these widgets. The widget below shows an example of
+          # usage, where we add a digital clock, setting 12h time and first day
+          # of the week to sunday.
+          {
+            digitalClock = {
+              calendar.firstDayOfWeek = "sunday";
+              time.format = "12h";
+            };
+          }
         ];
         hiding = "autohide";
       }
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -1,13 +1,13 @@
-{
-  config,
-  lib,
-  ...
-} @ args: let
+{ config
+, lib
+, ...
+} @ args:
+let
   cfg = config.programs.plasma;
 
   widgets = import ./widgets args;
 
-  panelType = lib.types.submodule ({config, ...}: {
+  panelType = lib.types.submodule ({ config, ... }: {
     options = {
       height = lib.mkOption {
         type = lib.types.int;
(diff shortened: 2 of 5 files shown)


magnouvean:
I have tested and it works great. I may even make some more widgets in the future. This was really clever, thank you.

PR #164, opened 2024-05-06
module(widgets): mass refactor, add system tray and battery

### LINKED ISSUE #126: Support for custom containments / system tray
Hello, thank you for your work!

I think it would be a nice feature to be able to configure the system tray the same way it is now possible with the rest of the panel applets. I looked at the module's code but I didn't find a way to do this, so of course you can close this if it's already possible 😅

In order to achieve this, it seems like there would just need to be support for containments of any type, which will probably also allow configuring desktop widgets or maybe even other things.

Unfortunately I'm new to nix so I didn't manage to understand how to change the code to make a PR 😅 But here are some thoughts on what I was looking for to maybe change, in case they help 🤔

* Maybe the `panel` module could be renamed/extended/aliased to/from `containment`
* Allow it to have `plugin=`{`org.kde.panel`/`org.kde.plasma.private.systemtray`/`org.kde.plasma.folder`}
* Specifically for the system tray, we would need to have access to the containments' ids while generating the configuration, as it is needed for `SystrayContainmentId`

Here is a minimal example of a system tray configuration to demonstrate the schema for this use case (if I have understood it correctly):

```ini
# Panel
[Containments][1]
plugin=org.kde.panel

## Panel widget #1
[Containments][1][Applets][1]
plugin=org.kde.plasma.icontasks

## Panel widget #2
[Containments][1][Applets][2]
plugin=org.kde.plasma.systemtray

[Containments][1][Applets][2][Configuration]
SystrayContainmentId=2

# System tray
[Containments][2]
plugin=org.kde.plasma.private.systemtray

## Tray icon #1
[Containments][2][Applets][1]
plugin=org.kde.plasma.devicenotifier

## Tray icon #2
[Containments][2][Applets][2]
plugin=org.kde.plasma.volume

[Containments][2][Applets][2][Shortcuts]
global=Meta+A
```

pluiedev:
Fixes #126 

Here's a list of changes:

- Most of the code that generates the JS required for adding and configuring the widgets are now moved to `modules/widgets/lib.nix`. Since system trays are also technically containments, and configuring things there is very similar to configuring a panel, it's worth refactoring it into code that can be shared for both.

- Every panel is now in its own scope, in order to avoid name collisions. As a direct consequence of that, all `var`s are replaced by `const`s, as there's now no need for redeclaring variables. Besides, `var` is a code smell anyways in modern JS due to its unpredictable scoping rules, and it's best to avoid it in new code.

- Config fields in the same group of the same widget will now share the `w.currentConfigGroup` and `w` variable assignments. Essentially, instead of this:

  ```js
  var w = panelWidgets["org.kde.plasma.systemmonitor"];
  w.currentConfigGroup = ["Appearance"];
  w.writeConfig("title", "CPU Usage");

  var w = panelWidgets["org.kde.plasma.systemmonitor"];
  w.currentConfigGroup = ["Sensors"];
  w.writeConfig("highPrioritySensorIds", '["cpu/all/usage"]');

  var w = panelWidgets["org.kde.plasma.systemmonitor"];
  w.currentConfigGroup = ["Sensors"];
  w.writeConfig("lowPrioritySensorIds", '["cpu/all/temperature"]');

  var w = panelWidgets["org.kde.plasma.systemmonitor"];
  w.currentConfigGroup = ["Sensors"];
  w.writeConfig("totalSensors", '["cpu/all/usage"]');
  ```

  We now generate:

  ```js
  const w = panelWidgets["org.kde.plasma.systemmonitor"];
  w.currentConfigGroup = ["Appearance"];
  w.writeConfig("title", "CPU Usage");

  w.currentConfigGroup = ["Sensors"];
  w.writeConfig("highPrioritySensorIds", "[\"cpu/all/usage\""]);
  w.writeConfig("lowPrioritySensorIds", "[\"cpu/all/temperature\""]);
  w.writeConfig("totalSensors", "[\"cpu/all/usage\"]");
  ```

  This _massively_ cuts down generated code, and is much cleaner in general.

- Every setting on every widget is now nullable. Setting anything to null basically makes it unset, which allows us to eliminate setting a default on the Nix side, so that upstream default changes won't affect us.
- An `extraConfig` field has been added for raw widgets, that takes a lambda/anonymous function that accepts one argument — the widget to be configured. I figure this allows more room for the user to name the widget whatever they want, and not rely on hardcoded variable names.

  This is _required_ to make system trays work, because configuring them is... quite peculiar, to say the least.

- I've added a link to the config/main.xml file that defines each widget's config options, which is a much more reliable source than KDE docs.

- The widget system now makes use of native module system features (such as the `apply` field for `mkOption`), in order to make converters less of an eyesore, make Nix report better errors, and make the system more robust against erroneous user input in general.

- And of course, you can now customize system trays to your liking, without knowing what eldritch horror lies beneath to make it work.

- Also, battery indicators!

### DIFF
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -60,18 +60,23 @@
           # If no configuration is needed, specifying only the name of the
           # widget will add them with the default configuration.
           "org.kde.plasma.marginsseparator"
-          "org.kde.plasma.systemtray"
           # If you need configuration for your widget, instead of specifying the
           # the keys and values directly using the config attribute as shown
           # above, plasma-manager also provides some higher-level interfaces for
           # configuring the widgets. See modules/widgets for supported widgets
-          # and options for these widgets. The widget below shows an example of
-          # usage, where we add a digital clock, setting 12h time and first day
-          # of the week to sunday.
+          # and options for these widgets. The widgets below shows two examples
+          # of usage, one where we add a digital clock, setting 12h time and
+          # first day of the week to sunday and another adding a systray with
+          # all systray-entries shown.
           {
             digitalClock = {
               calendar.firstDayOfWeek = "sunday";
               time.format = "12h";
+            };
+          }
+          {
+            systemTray = {
+              items.showAll = true;
             };
           }
         ];
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -67,7 +67,7 @@
           # and options for these widgets. The widgets below shows two examples
           # of usage, one where we add a digital clock, setting 12h time and
           # first day of the week to sunday and another adding a systray with
-          # all systray-entries shown.
+          # some modifications in which entries to show.
           {
             digitalClock = {
               calendar.firstDayOfWeek = "sunday";
@@ -75,8 +75,17 @@
             };
           }
           {
-            systemTray = {
-              items.showAll = true;
+            systemTray.items = {
+              # We explicitly show bluetooth and battery
+              shown = [
+                "org.kde.plasma.battery"
+                "org.kde.plasma.bluetooth"
+              ];
+              # And explicitly hide networkmanagement and volume
+              hidden = [
+                "org.kde.plasma.networkmanagement"
+                "org.kde.plasma.volume"
+              ];
             };
           }
         ];
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -76,7 +76,7 @@ let
       };
       floating = lib.mkEnableOption "Enable or disable floating style (plasma 6 only).";
       widgets = lib.mkOption {
-        type = with lib.types; listOf (either str widgets.type);
+        type = lib.types.listOf widgets.type;
         default = [
           "org.kde.plasma.kickoff"
           "org.kde.plasma.pager"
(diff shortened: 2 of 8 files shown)


pluiedev:
I also realized that users would be able to do something patently stupid like:
```nix
{
  widgets = [
    {
      systemTray.items.configs = {
        systemTray.items.configs = {
          systemTray.items.configs = {
            battery.showPercentage = true;
          };
        };
      };
    }
  ];
}
```

Should you do this? No. Should you be able to do this? Yes.

magnouvean:
Nice. It's quite a big pr, but hopefully I can test it this weekend.

magnouvean:
Works great as usual :). Will merge when checks are done. Thanks a lot once again!

lgoette:
Hey, thanks for the contribution. Works great :D
Is it possible to set `popupHeight` and `popupWidth` of the systemtray? I did not find options for these settings.

pluiedev:
> Hey, thanks for the contribution. Works great :D Is it possible to set `popupHeight` and `popupWidth` of the systemtray? I did not find options for these settings.

Give it a try - it might not work 'cause I didn't set the `freeformType`, though

lgoette:
I am not that familiar with how this works, sorry. But while `iconSpacing` etc is set in the `[Containments][systrayID][General]` location, `popupHeight` is set in the `[Containments][systrayID]` location. Is there an option to access this part of the config?

PR #181, opened 2024-06-01
system-monitor widget: Add label parameter for sensors

gkaklas:
Hello, thank you everyone for your work!

This is a small change to allow adding a label for a system monitor panel widget.

I did it by copying the code pattern currently used for colors (I'm sure it could be done better by creating re-usable modules etc for lines 40 and 101, but I'm new to Nix so I wouldn't know how to do it yet :sweat_smile: )

### DIFF
--- a/modules/widgets/system-monitor.nix
+++ b/modules/widgets/system-monitor.nix
@@ -31,10 +31,20 @@ let
   toColorKV =
     { name
     , color
+    , label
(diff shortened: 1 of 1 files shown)


magnouvean:
You solution looks just fine :). Works good, so I'll merge when checks are done. Thanks.

PR #182, opened 2024-06-01
Change lib.mergeAttrsList to lib.attrsets.mergeAttrsList

Antetokounpo:
Hello,

This is a tiny change to use `lib.attrsets.mergeAttrsList` instead of `lib.mergeAttrsList`, because the latter doesn't seem to exist in nixpkgs-23.11. 

Thanks for all the work on this project !

### DIFF
--- a/modules/widgets/default.nix
+++ b/modules/widgets/default.nix
@@ -4,7 +4,7 @@ let
     widgets = self;
   };
 
-  sources = lib.mergeAttrsList (map (s: import s args') [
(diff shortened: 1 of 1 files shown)


magnouvean:
Ah wasn't aware of that (all my testing is done on unstable). I think we can maybe merge into `trunk` and then I'll rebase for `plasma-5` branch soon if that's OK?

Antetokounpo:
Yep LGTM

magnouvean:
Nice. I'll try to get it into `plasma-5` in not too long.

PR #185, opened 2024-06-04
Never overwrite files in `overrideConfigExclude`

### LINKED ISSUE #184: `overrideConfigExclude` not working as expected
I've been using this configuration for a while now, so that my `kcminputrc` is not reset:
```nix
programs.plasma = {
  enable = true;
  overrideConfig = true;
  overrideConfigExclude = [ "kcminputrc" ];
};
```
This worked previously, but this recent change causes `programs.plasma.configFile.kcminputrc` to exist:
https://github.com/pjones/plasma-manager/blob/7c816ab287574c317ac586900718aa7ab85f2f17/modules/workspace.nix#L181

This seems to cause `write_config.py` to overwrite `kcminputrc` with nothing, erasing the settings there.

Because this file often contains convoluted hardware-specific names (e.g. `[Libinput][1267][13016][ASUE120C:00 04F3:32D8 Touchpad]`), manually enabling persistence for specific keys is difficult. 

Is there a way to ensure that this file is not erased when using `overrideConfig`?

InternetUnexplorer:
Fixes #184.

I made an effort to make this fix as simple as possible, but there is one change in behavior with this approach: `configFile`s that are not in `overrideConfigFiles` or `overrideConfigExclude` will no longer be overwritten when `overrideConfig` is enabled, whereas previously they would be.

I can change this if it is a problem, just let me know. The documentation seems vague on whether or not the current behavior is intentional, so I wasn't sure.

Also, I haven't done much testing yet; the diff is pretty small so I'm not too worried, but a sanity check would be appreciated. Thanks :)

### DIFF
--- a/script/write_config.py
+++ b/script/write_config.py
@@ -171,15 +171,10 @@ class KConfManager:
                     )
                 elif value["persistent"]:
                     base_msg = f'Plasma-manager: Persistency enabled for key "{key}" in group "{group}" in configfile "{self.filepath}"'
-                    # We don't allow persistency when not using overrideConfig,
-                    # the value is set, immutability is enabled, or when
-                    # shell-expansion is enabled.
-                    if not self.override_config:
-                        raise Exception(
-                            f"{base_msg} when overrideConfig is disabled (or using overrideConfigExclude). "
-                            "Persistency without using overrideConfig is not supported"
-                        )
-                    elif value["value"] is not None:
+                    # We don't allow persistency when the value is set,
+                    # immutability is enabled, or when shell-expansion is
+                    # enabled.
+                    if value["value"] is not None:
                         raise Exception(
                             f"{base_msg} with non-null value \"{value['value']}\". "
                             "A value cannot be given when persistency is enabled"
--- a/script/write_config.py
+++ b/script/write_config.py
@@ -176,7 +176,7 @@ class KConfManager:
                     # shell-expansion is enabled.
                     if not self.override_config:
                         raise Exception(
-                            f"{base_msg} when overrideConfig is disabled. "
+                            f"{base_msg} when overrideConfig is disabled (or using overrideConfigExclude). "
                             "Persistency without using overrideConfig is not supported"
                         )
                     elif value["value"] is not None:
@@ -311,9 +311,9 @@ def remove_config_files(d: Dict, reset_files: Set):
                 os.remove(file_to_del)
 
 
-def write_configs(d: Dict, override_config: bool):
+def write_configs(d: Dict, reset_files: Set):
     for filepath, c in d.items():
-        config = KConfManager(filepath, c, override_config)
+        config = KConfManager(filepath, c, filepath in reset_files)
         config.run()
         config.save()
 
@@ -331,7 +331,6 @@ def main():
     # We send in "true" as the second argument if overrideConfig is enabled in
     # plasma-manager.
     override_config = bool(sys.argv[2])
-    # os.system(f"echo '{sys.argv[2]}' > /home/user1/Downloads/test")
     # The files to be reset when we have overrideConfig enabled.
     oc_reset_files = set(sys.argv[3].split(" "))
 
@@ -340,7 +339,7 @@ def main():
     # not configured through plasma-manager.
     if override_config:
         remove_config_files(d, oc_reset_files)
-    write_configs(d, override_config)
+    write_configs(d, oc_reset_files if override_config else {})
 
 
 if __name__ == "__main__":


magnouvean:
You are correct that files not in `overrideConfigFiles` or `overrideConfigExclude` should not be overwritten, though it should still be possible to write keys to files which isn't in either of these (I guess I'll test that this works). If this happened previously, the reason for them being so previously was due to a bug. I'll check if this works, it looks quite elegant on first glance, and if all is OK I'll merge right away.

magnouvean:
Looks great for the most part. There is one problem though which is if persistency is enabled for a file which is added to `overrideConfigExclude` it crashes the whole configuration-writing script. This is particularly a problem since we enable this by default for some options. The incompatibility of not using `overrideConfig` and `persistence` is by design, since it doesn't make sense then. The problem seems to be that `KConfManager` after this PR reads the `overrideConfig` on a per-file basis, and if the file then has persistent keys it simply will crash. Finding a fix for this shouldn't be that hard though.

InternetUnexplorer:
Oh, good point. Wouldn't it be better to do these checks during eval in general? I can definitely implement that if that's how we want to go ahead. Looking at the code, I was thinking:
- Check for persistence/`overrideConfig(Exclude)` issues during eval instead of activation
- Simplify activation script a bit based on above change
- Clarify wording of `overrideConfig`/`overrideConfigFiles`/`overrideConfigExclude` options
- Revisit how arguments to activation script are escaped (unrelated, I just noticed it when making this change, can be done later)

Does this seem like an OK plan? If so, I'll go ahead and make these changes when I get a chance :)

InternetUnexplorer:
On second thought, I'm not sure I understand how `persistent` and `overrideConfig` are supposed to interact in general; wouldn't the activation script delete the entire file anyway because it's part of `oc_reset_files`?

It seems that with the current implementation `persistent` is only usable with files that are not in `overrideConfigFiles` or are in `overrideConfigExclude`, so maybe I need to rethink how this should work.

magnouvean:
Yes during eval would be better and was how I started off initially, but after trying for way too long and still getting rather incomprehensible errors with nested attrsets I remember I just gave up and went for the next best thing, but if you have the skills to implement this that would be wonderful.

`overrideConfig` is already explained a bit in the README, but I agree that we could do with better wording of the others. We may also want to add an option for adding additional files to the `overrideConfigFile`. In the future we could maybe even move all these options to one attrset, i.e. `overrideConfig.enable` `overrideConfig.exclude` and so on, but that is a bit much for this pr I reckon.

The above plan looks great!

magnouvean:
`overrideConfig` was originally designed to delete all the files in `overrideConfigFiles` (minus `overrideConfigExclude`), before writing the settings specified by `plasma-manager` to the file and as such plasma would just insert the default keys for the settings not set by `plasma-manager`. Later we added persistence so that keys marked as persistent was not deleted on activation, but left unchanged (while all the other keys were still deleted). Persistence doesn't make sense when we don't use `overrideConfig` though, since then no content of the config-files are deleted (though some keys may be overwritten).

InternetUnexplorer:
I still don't understand how persistence works as-is though; if `overrideConfig` is enabled, the entire file will be deleted by `remove_config_files` before `write_configs` is even called. Am I missing something?

magnouvean:
I believe the persistent keys and values will be saved in the class before deleting the file. It will then be deleted and then written again when writing to the file.

magnouvean:
Actually you may be right. There has been some changes since I last wrote this. I know persistence does work now though.

InternetUnexplorer:
The `KConfManager`s are created by `write_configs`, which is called after `remove_config_files`, so I don't think this works as expected, unless I am missing something:
https://github.com/pjones/plasma-manager/blob/0b3f430fe12d92c389aaf9cda27aa6678c780b1d/script/write_config.py#L341-L343
https://github.com/pjones/plasma-manager/blob/0b3f430fe12d92c389aaf9cda27aa6678c780b1d/script/write_config.py#L314-L318

Edit: persistence may work as expected currently, I just want to make sure I understand what the expected behavior is so I don't break it while making changes :)

magnouvean:
I think I see it now. `remove_config_files` removes only files which does not have any keys in the plasma-manager config (hence if it has persistent keys the file won't be removed). This is why it takes `d` (the plasma-manager configuration, formatted as a json) as an argument in addition to the files to delete.

InternetUnexplorer:
Oh! I see it now too, thanks for explaining :)

I think I need to think a bit about the best way to approach this; I'll revisit and try a few things when I get a chance.

Thanks a bunch!

magnouvean:
Sounds good. Thx for the effort :)

InternetUnexplorer:
So, it seems like are a bunch of different cases that need to be handled here. After brainstorming different ways to fix this, I'm wondering if the best way forward would be to only make the changes necessary to fix the immediate issue (activation error when persistence is enabled) and revisit everything else later.

For this, I'm wondering if the error in question is even necessary; sure, setting `persistent` when `overrideConfig(Exclude)` is set has no effect, because _every_ key is essentially persistent, but that doesn't seem harmful to me.

I can think of several situations where allowing this would be useful, including making it easier to disable `overrideConfig` temporarily to test things. This seems similar to how NixOS generally allows `foo.option` to be set even when `foo.enable` is `false`.

I haven't thought too much about the other checks there, but based on the above just removing the check seems like the simplest fix to the current issue, and the risk seems low (worst case some people will have `persistent` set for some keys without needing to).

Then, if this fixes the issue for now, maybe I can revisit this later with a different PR to tidy it up, try moving the assertions to eval time, etc. There are just a lot of cases to consider and I want to minimize the chances I break things right now.

magnouvean:
I agree it was never harmful, it just didn't make any difference so it wasn't allowed. I do agree with you though, especially regarding `foo.option` with `foo.enable`, as this is pretty much the same thing. It also makes it so that disabling `overrideConfig` doesn't make it so that one must manually disable all persistent keys throughout the config. I think this solution is fine for now anyway. I'll just check if this change calls for some adjustments elsewhere in `plasma-manager` (documentation etc).

magnouvean:
OK  so I have now tested that the following works:
- Files added to `overrideConfigExclude` doesn't get overwritten just because there is a persistent key in the file
- When `overrideConfig` is disabled no file gets overwritten
- When using `overrideConfig` the configuration outside of `plasma-manager` is removed on activation
- When using `overrideConfig` and we have enabled persistency the key does not get deleted

I think that is pretty much everything working. I'll merge right away and if we have missed something the fix should be pretty straight forward.

InternetUnexplorer:
Yay, thanks so much for your help with this :)

PR #186, opened 2024-06-04
add ui colorscheme support for kate and konsole

Asqiir:
This PR adds 2 new options:
* `programs.kate.ui.colorScheme` and
* `programs.konsole.ui.colorScheme`

and adds the values into `konsolerc`/`katerc`.

The rest of this PR that looks so big is some reordering in the kate file (sry).

### DIFF
--- a/modules/apps/konsole.nix
+++ b/modules/apps/konsole.nix
@@ -110,6 +110,16 @@ in
       '';
     };
 
+    ui.colorScheme = lib.mkOption {
+      type = with lib.types; nullOr str;
+      default = null;
+      example = "Krita dark orange";
+      description = ''
+        The colour scheme of the UI. Leave this setting at `null` in order to
+        not override the systems default scheme for for this application.
+      '';
+    };
+
     extraConfig = lib.mkOption {
       type = with lib.types; nullOr (attrsOf (attrsOf (basicSettingsType)));
       default = null;
@@ -132,6 +142,9 @@ in
             (lib.mapAttrs (keyName: keyAttrs: { value = keyAttrs; }) groupAttrs))
           cfg.extraConfig)
       )
+      {
+        "UiSettings"."ColorScheme".value = cfg.ui.colorScheme;
+      }
     ];
 
     xdg.dataFile = lib.mkMerge [
--- a/modules/apps/konsole.nix
+++ b/modules/apps/konsole.nix
@@ -143,7 +143,12 @@ in
           cfg.extraConfig)
       )
       {
-        "UiSettings"."ColorScheme".value = lib.mkIf (cfg.ui.colorScheme != null) cfg.ui.colorScheme;
+        "UiSettings"."ColorScheme" = lib.mkIf (cfg.ui.colorScheme != null) {
+          value = cfg.ui.colorScheme;
+          # if this wasn't immutable, it currently doesn't work when overrideConfig is activated
+          # see also: discussion at https://github.com/pjones/plasma-manager/pull/186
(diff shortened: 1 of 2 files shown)


magnouvean:
Seems to work fine for kate at the very least. Where is it possible to see the result for konsole? I have it applied but I'm not sure I see any difference (so I don't know if it's working or not).

Asqiir:
When you click on `Settings>>Window Color Scheme` you see all the options in the drop down menu. The currently chosen should change when changing `programs.console.ui.colorScheme` (and restarting konsole afterwards).

It should be most obvious when picking a dark theme if you typically use light themes ("Breeze Dark") or the other way around ("Breeze Light").

magnouvean:
Hmm, for me there is no `Settings>>Window Color Scheme` in konsole. For kate i see this though (and this works great for kate). I'm also not seeing any changes in konsole after setting this option, and I'm not seeing any option for setting ui colorscheme in the settings gui either, so I just need to make sure it makes sense to include it in `plasma-manager`.

Asqiir:
Right click anywhere in the console >> Menu (the lowest option) >> More (the lowest option) >> Settings >> Window Color Scheme

Asqiir:
It's easier to reach if one uses some of the toolbars.
And this setting (mostly) changes the colors of these toolbars.

magnouvean:
You are right, I missed that. The changes are rather minor but I see why you would want to set this :). This looks good. I'll merge this by the evening I probably.

modules/apps/konsole.nix
```diff
@@ -132,6 +142,9 @@ in
             (lib.mapAttrs (keyName: keyAttrs: { value = keyAttrs; }) groupAttrs))
           cfg.extraConfig)
       )
+      {
+        "UiSettings"."ColorScheme".value = lib.mkIf (cfg.ui.colorScheme != null) cfg.ui.colorScheme;
```
magnouvean: I need immutability here for this to not simply be overwritten when starting konsole for some reason.

magnouvean: Once this is solved I can merge :)

Asqiir: Sure, I will make that immutable.

Some more notes:

* I tried to run `home-manager switch` from inside konsole (checking my working theory, that konsole writes the current theme into the konsolerc upon closing the window), but that's not it.
* I do not have the same bug, so it is introduced somewhere where our systems differ. (I do not use overrideConfig, do you?)
* I would like to write down somewhere, where it's useful for future bug hunting. (Maybe an issue? Or a "known-weird-behaviour.md"?)

magnouvean: Maybe it has something to do with using `overrideConfig`. I'm also using konsole version "24.05.0". The behavior is illustrated if I edit `konsolerc` manually by adding (initially having the default colorscheme):
```nix
[UiSettings]
ColorScheme=Krita darker
```
and then launch konsole, this will have changed to
```nix
[UiSettings]
ColorScheme=
```
(while some new lines will have been added). Setting the key to immutable will not have this effect.

Regarding writing it down, I reckon adding a comment in the code right above the `immutability = true;` part suffices, maybe something along the lines of "Immutability seems to be needed when using overrideConfig (see https://github.com/pjones/plasma-manager/pull/186 for details)" (if `overrideConfig` is indeed the issue)

Asqiir: I changed something, if this doesn't fix it idk either.

magnouvean:
I'm happy to merge this now. Thanks for the pr :)

PR #190, opened 2024-06-05
add bracket config options for kate

Asqiir:
This adds 5 new options for kate:
* which characters are brackets
* whether closing brackets are automatically added when typing an opening bracket
*  whether the area between 2 brackets shall be highlighted when the cursor is next to one of the brackets
* whether to show the beginning (2 lines) of the opening bracket should be in a small, floating window, when the opening bracket is outside the current screen area and the cursor is next to the closing bracket
* whether the matching bracket should quickly flash when the cursor moves next to the other bracket

### DIFF
--- a/modules/apps/kate/default.nix
+++ b/modules/apps/kate/default.nix
@@ -232,31 +232,31 @@ in
   #     WRITING THE KATERC
   config.programs.plasma.configFile."katerc" = lib.mkIf cfg.enable {
     "KTextEditor Document" = {
-      "Auto Detect Indent".value = cfg.editor.indent.autodetect;
-      "Indentation Width".value = cfg.editor.indent.width;
+      "Auto Detect Indent" = cfg.editor.indent.autodetect;
+      "Indentation Width" = cfg.editor.indent.width;
       "Tab Handling" = (tabHandlingMode cfg.editor.indent);
-      "Tab Width".value = cfg.editor.tabWidth;
-      "Keep Extra Spaces".value = cfg.editor.indent.keepExtraSpaces;
-      "ReplaceTabsDyn".value = cfg.editor.indent.replaceWithSpaces;
+      "Tab Width" = cfg.editor.tabWidth;
+      "Keep Extra Spaces" = cfg.editor.indent.keepExtraSpaces;
+      "ReplaceTabsDyn" = cfg.editor.indent.replaceWithSpaces;
     };
 
     "KTextEditor Renderer" = {
-      "Show Indentation Lines".value = cfg.editor.indent.showLines;
+      "Show Indentation Lines" = cfg.editor.indent.showLines;
 
-      "Animate Bracket Matching".value = cfg.editor.brackets.flashMatching;
+      "Animate Bracket Matching" = cfg.editor.brackets.flashMatching;
 
       # Do pick the theme if the user chose one,
       # Do not touch the theme settings otherwise
-      "Auto Color Theme Selection".value = lib.mkIf (cfg.editor.theme.name != "") false;
-      "Color Theme".value = lib.mkIf (cfg.editor.theme.name != "") cfg.editor.theme.name;
+      "Auto Color Theme Selection" = lib.mkIf (cfg.editor.theme.name != "") false;
+      "Color Theme" = lib.mkIf (cfg.editor.theme.name != "") cfg.editor.theme.name;
     };
 
     "KTextEditor View" = {
-      "Chars To Enclose Selection".value = cfg.editor.brackets.characters;
-      "Bracket Match Preview".value = cfg.editor.brackets.highlightMatching;
-      "Auto Brackets".value = cfg.editor.brackets.automaticallyAddClosing;
+      "Chars To Enclose Selection" = cfg.editor.brackets.characters;
+      "Bracket Match Preview" = cfg.editor.brackets.highlightMatching;
+      "Auto Brackets" = cfg.editor.brackets.automaticallyAddClosing;
     };
 
-    "UiSettings"."ColorScheme".value = lib.mkIf (cfg.ui.colorScheme != null) cfg.ui.colorScheme;
+    "UiSettings"."ColorScheme" = lib.mkIf (cfg.ui.colorScheme != null) cfg.ui.colorScheme;
   };
 }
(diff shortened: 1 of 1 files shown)


Asqiir:
@magnouvean let me know whether you need immutability again for this to work

modules/apps/kate/default.nix
```diff
@@ -196,25 +196,65 @@ in
...
+  #     BRACKETS
+  options.programs.kate.editor.brackets = {
+    characters = lib.mkOption {
+      type = lib.types.str;
+      default = "<>(){}[]'\"\`";
+      example = "<>(){}[]'\"\`*_~";
+      description = "Which characters kate will treat as brackets.";
+    };
+    automaticallyAddClosing = lib.mkEnableOption {
+      description = "that a closing bracket is automatically inserted upon typing the opening";
```
magnouvean: Capitalization and punctuation

modules/apps/kate/default.nix
```diff
@@ -196,25 +196,65 @@ in
...
+    };
+    highlightRangeBetween = lib.mkEnableOption {
+      description = ''
+        that the range between the brackets is highlighted when the cursor is adjacent
```
magnouvean: Same here

modules/apps/kate/default.nix
```diff
@@ -196,25 +196,65 @@ in
...
+        to any of the brackets
+      '';
+    };
+    highlightMatching = lib.mkEnableOption {
+      description = ''
+        that the first two lines of the matching opening bracket are shown in a small floating window
```
magnouvean: And here

modules/apps/kate/default.nix
```diff
@@ -196,25 +196,65 @@ in
...
+        when the cursor is next to the closing bracket and the opening bracket is not in the visible
+        area
+      '';
+    };
+    flashMatching = lib.mkEnableOption {
+      description = ''
+        that the matching bracket flashes when the cursor moves adjacent to the other
```
magnouvean: And here

modules/apps/kate/default.nix
```diff
@@ -196,25 +196,65 @@ in
...
+  };
+
   # ==================================
   #     WRITING THE KATERC
   config.programs.plasma.configFile."katerc" = lib.mkIf cfg.enable {
     "KTextEditor Document" = {
-      "Auto Detect Indent" = cfg.editor.indent.autodetect;
-      "Indentation Width" = cfg.editor.indent.width;
+      "Auto Detect Indent".value = cfg.editor.indent.autodetect;
+      "Indentation Width".value = cfg.editor.indent.width;
```
magnouvean: Can drop `.value`. It doesn't matter, but after they were made optional the rest of the project doesn't use them unless you need to set `immutability` or something similar for the key.

modules/apps/kate/default.nix
```diff
@@ -196,25 +196,65 @@ in
...
       "Tab Handling" = (tabHandlingMode cfg.editor.indent);
-      "Tab Width" = cfg.editor.tabWidth;
-      "Keep Extra Spaces" = cfg.editor.indent.keepExtraSpaces;
-      "ReplaceTabsDyn" = cfg.editor.indent.replaceWithSpaces;
+      "Tab Width".value = cfg.editor.tabWidth;
+      "Keep Extra Spaces".value = cfg.editor.indent.keepExtraSpaces;
+      "ReplaceTabsDyn".value = cfg.editor.indent.replaceWithSpaces;
```
magnouvean: Same applies here

modules/apps/kate/default.nix
```diff
@@ -196,25 +196,65 @@ in
...
     };
 
     "KTextEditor Renderer" = {
-      "Show Indentation Lines" = cfg.editor.indent.showLines;
+      "Show Indentation Lines".value = cfg.editor.indent.showLines;
+
+      "Animate Bracket Matching".value = cfg.editor.brackets.flashMatching;
```
magnouvean: And here

modules/apps/kate/default.nix
```diff
@@ -196,25 +196,65 @@ in
...
-      "Auto Color Theme Selection" = lib.mkIf (cfg.editor.theme.name != "") false;
-      "Color Theme" = lib.mkIf (cfg.editor.theme.name != "") cfg.editor.theme.name;
+      "Auto Color Theme Selection".value = lib.mkIf (cfg.editor.theme.name != "") false;
+      "Color Theme".value = lib.mkIf (cfg.editor.theme.name != "") cfg.editor.theme.name;
+    };
+
+    "KTextEditor View" = {
+      "Chars To Enclose Selection".value = cfg.editor.brackets.characters;
+      "Bracket Match Preview".value = cfg.editor.brackets.highlightMatching;
+      "Auto Brackets".value = cfg.editor.brackets.automaticallyAddClosing;
```
magnouvean: And here

magnouvean (commented):
Seems to work fine for me without any immutability. Probably just some konsole problem. Looks good once the suggestions have been addressed.

modules/apps/kate/default.nix
Asqiir: ~~What do you mean?
`programs.kate.editor.brackets.characters` is defined via `mkOption`, which means, I have to start (by capitalizing) and end (by punctuation) the sentence myself. Which isn't the case for all the other options, that are defined with `mkEnableOption`.~~

modules/apps/kate/default.nix
Asqiir: Oh. I misunderstood what the default description of `mkEnableOption` does. Will fix.

magnouvean: I was mainly talking about the punctuation and capitalization in the sentence, but you also don't need the description field I don't think. For example `lib.mkEnableOption ''That the range between the brackets is highlighted when the cursor is adjacent.''` would be better (though maybe written slightly differently as the formulation seems a little weird to me).

Asqiir: yes, i thought it would automatically put "whether to enable ${my part}." into the description. I fixed it now.

magnouvean:
OK I'm ready to merge this now. One thing which may or may not be a problem is that `[]` appears as `\x5b\x5d` in `Chars To Enclose Selection`, though it may not be a problem (I don't really use kate so I don't know for sure). If it needs to be `[]` the functionality to make it so isn't yet available in `plasma-manager` (but hopefully this won't be the case for too long) so we can just make a PR when that happens if this actually is a problem. Thanks for this as always, and I'll merge the other PR right away as well :)

PR #193, opened 2024-06-07
Systray fixes

Sporif:
A couple of small fixes for the system tray module.


magnouvean:
Great, it works just as it should. I'll merge right away :)

PR #199, opened 2024-06-10
add module for configuring window rules

icewind1991:
Allow configuring kwin window rules through plasma-manager.

For future work it would be nice to have the `apply` options be a fully typed submodule with all the available options and their correct type instead of the current, more freeform, `attrsOf` to remove some of the guesswork from writing configurations. ~~But I was to lazy to do that for now.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ At the moment `plasma-manager` supports configuring the following:
 - Hotkeys (via the `hotkeys` module)
 - Panels (via the `panels` module)
 - Fonts (via the `fonts` module)
+- Window Rules (via the `window-rules` module)
 - KDE apps (via the `apps` module). In particular the following kde apps have
   modules in `plasma-manager`:
   - kate
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -112,7 +112,7 @@
             value = "dolphin";
             type = "substring";
           };
-          window-types = ["normal"];
+          window-types = [ "normal" ];
         };
         apply = {
           noborder = {
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -101,6 +101,28 @@
       }
     ];
 
+    window-rules = [
+      {
+        description = "Dolphin";
+        match = {
+          window-class = {
+            value = "dolphin";
+            type = "substring";
+          };
+          window-types = ["normal"];
+        };
+        apply = {
+          noborder = {
+            value = true;
+            apply = "force";
+          };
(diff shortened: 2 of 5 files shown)


magnouvean (commented):
I did test the example config, and after running `qdbus org.kde.KWin /KWin reconfigure` manually it did work well, but this had to be done manually, it did not happen automatically. With that said though, we could just drop the qdbus reloading part as the rest of `plasma-manager` a.t.m. needs fully logging out/in for things to work. We can worry about this in a later PR (if plasma provides the necessary tools for this to work smoothly across for most/all modules).

Otherwise it looks good, though if you could format with `nixpkgs-fmt` that would be great. Nice work on this :)

magnouvean:
I also see the checks fail, but that probably isn't due to this. I'll see if this is caused by something else.

icewind1991:
While using this for a while I ran into an issue where if the order of rules change or rules get removed, the news set of rules gets merged "randomly" with the old set of rules. Resulting in various unexpected window rules being create.

I've solved this locally by just setting `kwinrulesrc` to be deleted/overwritten.

Given that this currently already behaves as overwriting all "ui defined" window rules (because it has to set the list of configured rules). It might make sense to have it always delete `kwinrulesrc` regardless of whether `overrideConfig` is set. Though this could be considered unexpected behavior.

An alternative would be to extend the files module to allow specifying individual configuration blocks to be overwritten instead of merged.

magnouvean:
I see. I'll try to improve the interface so that we could enable deletion of some files even without `overrideConfig` (this list could start off by just having `kwinrulesrc`) and when I figure out a proper way of doing this it can be implemented here.

magnouvean:
OK so it should now be easy to make sure that `kwinrulesrc` is deleted upon activation even when not using `overrideConfig` (it should just be adding `kwinrulesrc` to https://github.com/pjones/plasma-manager/blob/trunk/modules/files.nix#L61). I would say that maybe this should only be added there in the case where `window-rules` is not empty (i.e. `builtins.length` of the list bigger than 0) as it then will not reset/remove this file in the case where the user doesn't configure window-rules from plasma-manager at all.

Fixing this and removing the qdbus part (I have never gotten much luck using `qdbus` in general for this project, especially for reloading `kwin`) should be enough before we can merge this.

icewind1991:
done

magnouvean:
I'm happy with this now. Thanks for this, this is really cool :)

PR #200, opened 2024-06-12
Fix font configuration for Plasma 5

### LINKED ISSUE #168: Font setting on Plasma 5 is broken
I am running Plasma 5 on NixOS 23.11. When I try to set the global monospace font with plasma-manager it does not work, and I get a errors logged when opening the fonts section in system-settings. I suspect the issue here may be applicable to all font settings.

This looks to be due to the fact font changes in Plasma 5/6 (Qt 5 vs 6) are not accounted for. In 6, there are 6 new font fields, and Plasma 5 does not like them existing in a font configuration string. Additionally the font weight enum is changed for Plasma 6. The enum values are completely different. See: [Qt5](https://doc.qt.io/qt-5/qfont.html) vs [Qt6](https://doc.qt.io/qt-6/qfont.html) docs.

I have not ready for PR (no cleanup of dead code) working fix in [my fork](https://github.com/luna-1024/plasma-manager/compare/9239c27..4235eba). Since all the new font options are at the end of the string, I just removed the last 6 items.

luna-1024:
Before the fix, setting the fonts results in a parsing error and the effective font reverting to a default instead.
When running a program, the error is logged by QFont::fromString() in Qt5. 

Refer to: [Qt 5 source](https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qfont.cpp?h=5.15#n2156)
Notice that it will only accept 10 items in the font string (excluding optional style name). 
Compare that to the Qt6 (for Plasma 6): [Qt 6 Source](https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qfont.cpp?h=6.7.1#n2182) we see that the unsupported fields we need to remove are the last 6 of them.

Additionally, the font weight enum changed in Qt6, so we have to fix it be the Qt5 scale.
Qt 5: https://doc.qt.io/qt-5/qfont.html#Weight-enum
Qt 6: https://doc.qt.io/qt-6/qfont.html#Weight-enum

Fixes #168

### DIFF
--- a/lib/qfont.nix
+++ b/lib/qfont.nix
@@ -20,15 +20,15 @@
 
     # QFont::Weight
     weight = {
-      thin = 100;
-      extraLight = 200;
-      light = 300;
-      normal = 400;
-      medium = 500;
-      demiBold = 600;
-      bold = 700;
-      extraBold = 800;
-      black = 900;
+      thin = 0;
+      extraLight = 12;
+      light = 25;
+      normal = 50;
+      medium = 57;
+      demiBold = 63;
+      bold = 75;
+      extraBold = 81;
+      black = 87;
     };
 
     # QFont::Style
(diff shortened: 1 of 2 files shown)


magnouvean:
Have you tested that this works? If so I am happy to merge right away :)

luna-1024:
I ran a system with Plasma 5, and tested if setting a global font (monospace type). 

Before this change it was broken:
- showed Noto Sans in place of the specified font in + logged errorin `systemsettings`
- KDE apps (konsole, kate) also log error and show built in(?) default font instead) 

The font setting worked properly after this change.

PR #202, opened 2024-06-13
Add option for cursor size

### LINKED ISSUE #198: Feature Request: Configure cursor size
Hi, I'm using this package alongside with stylix; I had to disable stylix because it's a bit buggy on KDE Plasma 6. I added in most of my stylix settings with plasma-manager but the one setting I couldn't find is the cursor size. This would be nice to have since we can already configure the theme.

magnouvean:
Closes #198.

### DIFF
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -32,13 +32,34 @@ let
     };
   };
 
+  cursorType = with lib.types; submodule {
+    options = {
+      cursorTheme = lib.mkOption {
+        type = nullOr str;
+        default = null;
+        example = "Breeze_Snow";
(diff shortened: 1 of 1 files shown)


modules/workspace.nix
```diff
@@ -32,13 +32,34 @@ let
     };
   };
 
+  cursorType = with lib.types; submodule {
+    options = {
+      cursorTheme = lib.mkOption {
```
bew, line 37: Why is this not called `theme` instead?
So the option path looks like `workspace.cursor.theme` similar to `workspace.cursor.size` 🤔

magnouvean: I can change it to that. I've never been great with naming variables and options tbh, but I see why that may be better.

PR #209, opened 2024-06-13
Fix date.format option for digital-clock

Sporif:


magnouvean:
While this does seem to work it wasn't really broken on my end before. Was there a specific configuration which didn't work and which this fixes?

Sporif:
Setting `date.format` to something other than `shortDate`, i.e `longDate` or `isoDate` won't do anything without this patch. `custom` still worked though.

magnouvean:
Ah, I only tested the `custom`, but you are correct. Looks good to merge then, thanks again.

PR #213, opened 2024-06-16
Add more kwin effects

magnouvean:
Adds quite a lot more of the kwin effects into `programs.plasma.kwin.effects`. A lot of the most important ones are now included, but some are still missing. A lot of the effects also support further customizations, and this hasn't been the focus of this PR, that can come later (though I added one additional option for the minimization effect).

### DIFF
--- a/modules/kwin.nix
+++ b/modules/kwin.nix
@@ -57,6 +57,12 @@ let
       (imap1 (i: v: (nameValuePair "Name_${builtins.toString i}" v)) names);
 in
 {
+  imports = [
+    (lib.mkRenamedOptionModule
+      [ "programs" "plasma" "kwin" "virtualDesktops" "animation" ]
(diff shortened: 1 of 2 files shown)


modules/kwin.nix
```diff
@@ -79,18 +85,94 @@ in
...
+          type = with types; nullOr ints.positive;
+          default = null;
+          example = 50;
+          description = ''
+            The duration of the minimization effect in milliseconds. Only
+            available when the minimization effect is magic lamp.
+          '';
+        };
+      };
+      wobblyWindows.enable = mkOption {
```
bew, line 112: I know wobbly effects has configurations, is that intentionality that it is not configurations here?

magnouvean: It's coming, but I focused on adding options for enabling first and I'll probably add options for configuring the different effects later.

PR #225, opened 2024-06-23
Add immutableByDefault option

### LINKED ISSUE #217: Add option `everythingImmutable` or `immutableByDefault`
I would like to have an option that makes *everything I set with plasma-manager* `immutable`.

Proposed name:
(1) `everythingImmutable`
(2) `forceImmutable`
(3) `immutableByDefault`

Proposed behaviour:
* every option for which a value but no mutability/persistency behaviour was specified: this is now immutable
* every option where `persistent` is specified: stays persistent
* when something is explicitly marked as `immutable = false` then (for 1 or 2) fail the build or (3) make not immutable. I don't have a strong intuition on that.

@magnouvean I think that goes with the spirit of your other proposal of changing the override behaviour. What do you think?

magnouvean:
This adds support for automatically marking all keys written by plasma-manager as immutable by default, including keys written by higher level modules. For this to be possible, immutability (and shellExpansion even though not strictly needed) and persistence have now become allowed for the same key. It's worth noting that immutability will in these cases not have any meaning, so for example:
```nix
group.key = {immutable = true; persistent = true;};
```
will result in the key `key` in group `group` being persistent, not `key[$i]`. I'm not completely settled on this choice, but it comes from the need of marking non-immutable keys as persistent as well. Typically the keys we want to mark as persistent are not immutable so it makes sense in this way (we basically never have to make immutable keys persistent). I would love other's thoughts on this as well. I have not done much testing, but it seems to work fine for me. Probably needs a little more testing to make sure everything works before I'll merge @Asqiir.
Closes #217.

### DIFF
--- a/lib/types.nix
+++ b/lib/types.nix
@@ -1,4 +1,4 @@
-{ lib, ... }:
+{ lib, config, ... }:
 let
   ##############################################################################
   # Types for storing settings.
@@ -13,7 +13,7 @@ let
       };
       immutable = lib.mkOption {
         type = bool;
-        default = false;
+        default = config.programs.plasma.immutableByDefault;
         description = ''
           Whether to make the key immutable. This corresponds to adding [$i] to
           the end of the key.
--- a/lib/writeconfig.nix
+++ b/lib/writeconfig.nix
@@ -23,9 +23,10 @@ let
           resetFilesList ++ [ "${config.xdg.dataHome}/plasma-manager/last_run_*" ]
         else
           resetFilesList);
+      immutableByDefault = (builtins.toString config.programs.plasma.immutableByDefault);
     in
     ''
-      ${writeConfigScript}/bin/write_config ${jsonFile} "${resetFilesStr}"
+      ${writeConfigScript}/bin/write_config ${jsonFile} "${resetFilesStr}" "${immutableByDefault}"
     '';
 in
 {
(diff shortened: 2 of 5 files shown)


Asqiir (approved):
I believe that this might bring us a lot of users later on that ask us "why does immutability not work??"

For the check that immutability and persistence aren't enabled at once: maybe only disable that check when `immutabilityByDefault` is enabled.

I believe that there might be a more elegant solution to achieve this, but this is already pretty good and pretty quick :)

magnouvean:
I like the suggestion, though the code implementation wouldn't be particularly elegant as we'd need to pass in extra arguments to the config-writing script and pass this into a lot of functions, which would clutter the code quite a bit. We can see if we get more bug reports following this change whether it's worth implementing, though it would only be applicable to cases where the user has enabled peristency manually together with immutability, which probably is quite niche of a use-case. Additionally, at least to me it doesn't really make sense that `group.key = { persistent = true; immutable = true; };` should make `key[$i]` persistent, so I kind of think the present solution makes just as much sense.
Great that it works! All the tests I've done have looked good, so I can probably merge rather soon :)

magnouvean:
Actually thinking about it, maybe it makes sense to check if the `immutable` setting is simply different from the default (so it'd error for persistent keys for non-immutable keys when immutableByDefault is enabled and for immutable keys when immutableByDefault is disabled, just like before). That I think is good enough of a solution to be worth the slight complication.

magnouvean:
I managed to implement this just fine I think. Will test some more and hopefully get this merged in a couple days.

PR #227, opened 2024-06-24
Remove the nix-community transfer snippet from the README file

HeitorAugustoLN:
Since the project has been successfully transferred to nix-community, we may now proceed to remove this snippet from the README file.

I think we could consider further additions to this section of the README file


magnouvean:
Yeah the section is a bit short, and the README probably needs some rewrites soon. We can do this in another PR so I'll merge this for now anyway :)

Issue #229, opened 2024-06-24
Implement widget-specific options and installation for third-party widgets.


HeitorAugustoLN:
## Objective

Create widget-specific options to facilitate the configuration and installation of third-party widgets in KDE Plasma environments. This enhancement aims to broaden user customization options within plasma-manager by making it easy to install and customize third-party widgets into the Plasma desktop environment, thereby enriching the user experience with plasma-manager.

## Things needed to do

For each widget that will be implemented, there are three things that need to be done:

1. Package the widget
2. Make it available in nixpkgs
3. Implement the widget-specific options in plasma-manager

### Widget packaging 

As of now, there is currently only one (that I know of) widget package available in nixpkgs for Plasma 6. And another one (that I know of) with a pull request open in nixpkgs. Them being:

#### In nixpkgs

- application-title-bar: nixos/nixpkgs#318103
- plasmusic-toolbar: nixos/nixpkgs#319671

> [!NOTE]
> Can have more that I don't know

#### Pull request open

> [!NOTE]
> Can have more that I don't know

#### Not packaged yet

> [!NOTE]
> Here we can add widgets asked by the users

#### Option

- Introduced in #233 

#### Supported packages:

- application-title-bar: #233

#### Done

- application-title-bar: #234 

#### Not started

magnouvean:
I reckon this is done now and that the further discussions can be held in the pr-s for future widgets

PR #234, opened 2024-06-28
Add application-title-bar widget-specific options

HeitorAugustoLN:
Implements third-party widget-specific options for application-title-bar as suggested in #229

### DIFF
--- a/modules/widgets/application-title-bar.nix
+++ b/modules/widgets/application-title-bar.nix
@@ -1,0 +1,446 @@
+{ lib, widgets, ... }:
+let
+  inherit (lib) mkOption types;
+  inherit (widgets.lib) mkBoolOption mkEnumOption;
+
+  convertHorizontalAlignment = horizontalAlignment: let
+    mappings = {
+      "left" = "1";
+      "right" = "2";
+      "center" = "4";
+      "justify" = "8";
+    };
+  in
+    mappings.${horizontalAlignment} or (throw "Invalid enum value: ${horizontalAlignment}");
+
+  convertVerticalAlignment = verticalAlignment: let
+    mappings = {
+      "top" = "1";
+      "center" = "128";
+      "bottom" = "64";
+      "baseline" = "256";
+    };
+  in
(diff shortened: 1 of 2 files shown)


modules/widgets/application-title-bar.nix
```diff
@@ -0,0 +1,446 @@
...
+          default = null;
+          description = "The horizontal alignment of the widget.";
+          apply = convertHorizontalAlignment;
+        };
+        verticalAlignment = mkOption {
+          type = types.nullOr (types.enum [ "top" "center" "bottom" "baseline" ]);
+          default = null;
+          description = "The vertical alignment of the widget.";
+          apply = convertVerticalAlignment;
+        };
```
magnouvean: These allow for `null` in the type (in fact it defaults to `null`), but it crashes when these options are `null`. Either allow for `null` values in `convertVericalAlignment`/`convertHorizontalAlignment` or don't allow them to take the value `null` at all (and set some default values)

magnouvean:
Other than that it looks OK for the most part I reckon. I don't really have time to test all these options though, but maybe I'll test a few of them tomorrow and see if they generally work. In the future we should maybe also add an option to add the widget automatically to `extraWidgets`, but the interface isn't really there yet so that can be done in another pr.

modules/widgets/application-title-bar.nix
HeitorAugustoLN: Fixed. I didn't notice it when I was writing it.

HeitorAugustoLN:
> Other than that it looks OK for the most part I reckon. I don't really have time to test all these options though, but maybe I'll test a few of them tomorrow and see if they generally work.

As far as I tested, everything should be working right now.

> In the future we should maybe also add an option to add the widget automatically to `extraWidgets`, but the interface isn't really there yet so that can be done in another pr.

That is a good idea, I could try doing it later.

magnouvean:
Nice that has done the trick. I'll merge right away then as this looks good. Great work!

HeitorAugustoLN:
Thanks!

PR #236, opened 2024-06-30
Automatically install third-party widget if it is present in a panel

HeitorAugustoLN:
As @magnouvean suggested, we could consider adding an option in the future to automatically add the widget to extraWidgets. I had this idea, which I believe could be even better, in my opinion.

I tested it in my configuration, and it is working flawlessly.

### Changes

- Adds `hasWidget` function for checking whether a specific widget, identified by its name `widgetName`, is present in any of the configured panels.
- Use `hasWidget` to automatically install a third-party widget if the user does not specify any in the extraWidgets option.

### DIFF
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -5,6 +5,7 @@
 } @ args:
 let
   cfg = config.programs.plasma;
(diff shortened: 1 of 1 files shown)


magnouvean:
Nice and simple solution to this problem. I'll merge right away.

PR #238, opened 2024-07-01
Add widget-specific options for Kickoff

HeitorAugustoLN:
Did not put some options such as: `favoritesPortedToKAstats`, `systemApplications`, `systemFavorites`. As probably it is not needed to be set. And there is not enough documentation on these options. Maybe i will look into it in the future, and maybe add it

### DIFF
--- a/modules/widgets/default.nix
+++ b/modules/widgets/default.nix
@@ -8,6 +8,7 @@ let
     ./application-title-bar.nix
     ./battery.nix
     ./digital-clock.nix
+    ./kickoff.nix
     ./system-monitor.nix
     ./system-tray.nix
   ]);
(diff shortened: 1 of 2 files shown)


HeitorAugustoLN:
Tested, it is working fine.

HeitorAugustoLN:
Only the `favorites` option doesn't work. The option is doing the job. And I don't even think it is our fault in the scripting. It never worked for me.

image
Sets everything fine. But the favorites doesn't update

image

magnouvean:
I'll try to get the favorites working myself (maybe it needs a restart of `plasma-plasmadesktop` maybe, or some other weird trick). If that doesn't work I think it would be better to drop it for now as having an option which doesn't work doesn't really make sense, even if the bug is on plasma desktop-scripts and not in `plasma-manager`.

It might take a couple days before I can check it out though.

HeitorAugustoLN:
I would say the bug is in plasma-desktop scripts. But if you can get it working, it would be nice. But if not, i will remove it from the options

magnouvean:
Yeah I don't know what's happening with the favorites either. We can find out at a later time. I reckon we just remove that and I can merge. Everything else looks fine.

HeitorAugustoLN:
> Yeah I don't know what's happening with the favorites either. We can find out at a later time. I reckon we just remove that and I can merge. Everything else looks fine.

Removed.

PR #244, opened 2024-07-03
Add module for Okular

Iorvethe:
This PR adds a new module for configuring Okular, based on the existing one for Kate. I tested it locally, and it seems to be working.

## Scope

I have, on purpose, added only a subset of all options for a first MR to keep it relatively small. Included are only those that are in my opinion most likely to be changed from the defaults. Feel free to suggest more, and I’ll happily add them!

## Future work

Features that would be interesting in the future:
- Add way to configure shortcuts. From what I understood, this is managed by KXMLGUI, and may also be interesting for other app modules.
- Add way to configure quick annotation tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ At the moment `plasma-manager` supports configuring the following:
   modules in `plasma-manager`:
   - kate
   - konsole
+  - okular
 
 Additionally there are more functionality than just listed above, and more
 functionality to come in the future!
--- a/modules/apps/default.nix
+++ b/modules/apps/default.nix
@@ -4,5 +4,6 @@
   imports = [
     ./konsole.nix
     ./kate
+    ./okular.nix
   ];
 }
--- a/modules/apps/okular.nix
+++ b/modules/apps/okular.nix
@@ -13,9 +13,7 @@ in with lib.types;
       example = "pkgs.libsForQt5.okular";
       extraDescription = ''
         Which okular package to install. Use `pkgs.libsForQt5.okular` in Plasma5 and
-        `pkgs.kdePackages.okular` in Plasma6. Use `null` if home-manager should not install okular
-        (use this if you want to manage the settings of this user of a system-wide okular
-        installation).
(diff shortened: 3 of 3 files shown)


magnouvean:
Nice, this looks promising. One thing which could be handy would be if the options can take `null` as a value (and takes this by default). Then for the options that are `null` we don't write any values at all (this can be done using `mkIf`-s). That way everything which isn't explicitly set by the user would be the default, and this default doesn't have to be set by `plasma-manager`. Especially for colors this makes sense I reckon as that may change between okular versions or different themes. I see that the `kate` module doesn't make use of this though, but the defaults there may be a little more standard/universal (and probably could/should be done eventually as well anyway).
Other than that it looks good. I will test in a couple of days and hopefully we can merge it soon.

Iorvethe:
Great, thanks for your review ! I will look into your suggestion regarding the `null` values as soon as possible. It’s indeed better to be more future-proof.

On another subject, I don’t know if it’s an issue with my configuration or in general, but for the `package` option, using `null` yields an error saying that it’s not of type `package`. The same error is reported for the `kate` module.

Do you also have this problem on your end?

magnouvean:
I think that's to be expected yes (I can't test right now as I don't have my computer available), since then nix would try to install `null` which isn't a valid package I presume. This doesn't need to be able to take `null` as a value though, only the other options it would be nice :)

Iorvethe:
I just pushed a commit with the modifications. It seems to work on a first glance. Hope everything is alright!

It makes total sense as to the `null` value for the `package`, it’s just that from what I understood a `null` value is supposed to tell plasma-manager that it should use the globally installed package, and just manage the configuration. I will remove it for the time being, though.

magnouvean:
Looks very good. I'll probably double check everything works some time in the weekend and let you know if there's any more adjustments needed, or if not merge it right away :)

magnouvean:
Seems to work great for what I've tested. I haven't managed to test everything, but if something is broken it's an easy fix. I'll merge once the CI has completed. Thanks for this and good work :)

PR #246, opened 2024-07-04
Standardize Enums Options for widget-specific options

HeitorAugustoLN:
This pull request standardizes the options for enums in the code; previously, there was no standard. The changes ensure consistency in the codebase.

### DIFF
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -73,7 +73,7 @@
           # some modifications in which entries to show.
           {
             digitalClock = {
-              calendar.firstDayOfWeek = "sunday";
+              calendar.firstDayOfWeek = "Sunday";
               time.format = "12h";
             };
           }
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -73,7 +73,7 @@
           # some modifications in which entries to show.
           {
             digitalClock = {
-              calendar.firstDayOfWeek = "Sunday";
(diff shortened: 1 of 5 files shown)


magnouvean:
I would maybe prefer not having the first letter capitalized as this is more consistent with how nix variables are named (unless there is some reason for capitalizing the first letter that makes more sense that I'm not seeing). Having consistency I think is nice though. This would probably break a lot of configs though, but I think that's fine as it's an easy fix to make.

HeitorAugustoLN:
> I would maybe prefer not having the first letter capitalized as this is more consistent with how nix variables are named (unless there is some reason for capitalizing the first letter that makes more sense that I'm not seeing). Having consistency I think is nice though. This would probably break a lot of configs though, but I think that's fine as it's an easy fix to make.


I hadn't considered that; this is probably better. I chose to capitalize the first letter of the enum values because it aligns with how KDE Plasma names their enum values, as demonstrated here:

image

magnouvean:
Hmm... I'm honestly not sure. I think for plasma consistency it's better with the current, but the nix code would look better with lowercase first letter. I'm slightly leaning towards lowercase since this is a high-level interface so consistency with kde plasma isn't THAT important. I guess it doesn't matter that much as long as we have consistency...

PR #248, opened 2024-07-06
Allow bools, ints and floats for widget configs and major refactoring

### LINKED ISSUE #247: [Feature]: Add option to control the data range of system monitor sensors
It could be good if the system monitor widget had the ability to set a non-automatic range. Currently, I couldn't find the option if it does exist.

magnouvean:
Allows setting for example:
```nix
{
  name = "org.kde.plasma.kickerdash";
  config.General.someval = 8;
}
```
in widgets, where we before the value had to be `"8"`. Same support is also added for bools and floats. Also refactored the widgets massively as the special types aren't needed anymore after this. Most of the pr consists of refactors as such.

Also adds `range` attributes to the system-monitor widget, so it should close #247.
Probably will need some more testing since it is a big pr, so if anyone have any time to test and see if everything works and report whether they run into any problems or not that would be great as I can merge quicker then.

### DIFF
--- a/modules/widgets/lib.nix
+++ b/modules/widgets/lib.nix
@@ -1,39 +1,35 @@
-{ lib, widgets, ... }:
+{ lib, ... }:
 let
   inherit (lib)
     optionalString
     concatMapStringsSep
     concatStringsSep
     mapAttrsToList
-    splitString
-    mkOption
-    types
-    ;
+    splitString;
+
+  configValueTypes = with lib.types; oneOf [ bool float int str ];
+  configValueType = with lib.types; nullOr (attrsOf (attrsOf (either configValueTypes (listOf configValueTypes))));
 
   # any value or null -> string -> string 
   # If value is null, returns the empty string, otherwise returns the provided string
   stringIfNotNull = v: optionalString (v != null);
 
-  # string -> string
-  # Wrap a string in double quotes.
-  wrapInQuotes = s: ''"${s}"'';
+  # Converts each datatype into an expression which can be parsed in JavaScript
+  valToJS = v: if (builtins.isString v) then ''"${v}"'' else if (builtins.isBool v) then (lib.boolToString v) else (builtins.toString v);
 
-  # list of strings -> string
-  # Converts a list of strings to a single string, that can be parsed as a string list in JavaScript
-  toJSStringList = values: ''[${concatMapStringsSep ", " wrapInQuotes values}]'';
+  # Converts a list of  to a single string, that can be parsed as a string list in JavaScript
+  toJSList = values: ''[${concatMapStringsSep ", " valToJS values}]'';
 
   setWidgetSettings = var: settings:
     let
       perConfig = group: key: value: ''${var}.writeConfig("${key}", ${
-        if builtins.isString value
-        then wrapInQuotes value
-        else if builtins.isList value
-        then toJSStringList value
-        else throw "widget config ${group}.${key} can only be string or string list, found ${builtins.typeOf value}"
-      });'';
+  if builtins.isList value
+  then toJSList value
+  else valToJS value
+    });'';
 
       perGroup = group: configs: ''
-        ${var}.currentConfigGroup = ${toJSStringList (splitString "/" group)};
+        ${var}.currentConfigGroup = ${toJSList (splitString "/" group)};
         ${concatStringsSep "\n" (mapAttrsToList (perConfig group) configs)}
       '';
     in
@@ -58,43 +54,11 @@ let
       const ${var} = {};
       ${lib.concatMapStringsSep "\n" addStmt ws}
     '';
-
-  boolToString' = b: if b == null then null else lib.boolToString b;
-
-  getEnum = es: e:
-    if e == null
-    then null
-    else
-      toString (
-        lib.lists.findFirstIndex
-          (x: x == e)
-          (throw "getEnum: nonexistent key ${e}! This is a bug!")
-          es
-      );
-
-  mkBoolOption = description: mkOption {
-    inherit description;
-
-    type = types.nullOr types.bool;
-    default = null;
-    example = true;
-    apply = widgets.lib.boolToString';
-  };
-
-  mkEnumOption = enum: mkOption {
-    type = types.nullOr (types.enum enum);
-    default = null;
-    apply = widgets.lib.getEnum enum;
-  };
 in
 {
   inherit
     stringIfNotNull
     setWidgetSettings
     addWidgetStmts
-    boolToString'
-    getEnum
-    mkBoolOption
-    mkEnumOption
-    ;
+    configValueType;
 }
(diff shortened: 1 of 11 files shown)


HeitorAugustoLN:
I haven't tested it yet, but I noticed that the functions `mkEnumOption` and `getEnum` have been removed. How will enum values be converted now? As they need to be converted to an integer that corresponds to the index of the enum value (AFAIK).

HeitorAugustoLN:
Or even when they are not, enum values are often converted to integers that correspond to the required values. For instance, consider the following example in `kickoff`:

```nix
showButtonsFor = mkEnumOption [ "power" "session" "custom" "powerAndSession" ] // {
  example = "powerAndSession";
  description = "Which actions should be displayed in the footer.";
};
```

This configuration sets `primaryActions` accordingly:

image

magnouvean:
They should be able to be specified using normal mkOption instead. The difference is that we shouldn't need to depend on some custom type.

magnouvean:
I can see that I may have misunderstood what the option does though, so I'll have to look into alternatives there. I think in general the custom type specification there was a little ugly and specific to some few use-cases. I'll think about some alternatives though and if I can't come up with any good ones I'll just bring it back through a refactored version (though maybe in `lib/types.nix` or something as this wouldn't be specific to widgets)

magnouvean:
Thinking about it I think we could make for example some attrset:
```nix
{
  power = 0;
  session = 1;
  custom = 2;
  powerAndSession = 3;
}
```
and then set the enum values to the keys and then in the apply for the option just extract the value using this attrset. An alternative is to keep using `mkEnumOption`, but making it a proper option so that we don't have the need for `//` syntax. In that case though I would prefer it instead of taking in just a list of allowed values that it instead takes in an attrset, which makes it clearer the values assigned to the different enum values (and also allows not starting from `0`).

I think what's good about the first of these two is that we don't depend on our own custom `lib` throughout the project, which will make it easier to get into coming from `nixpkgs` or `home-manager`, but it will be less compact code.

HeitorAugustoLN:
> They should be able to be specified using normal mkOption instead. The difference is that we shouldn't need to depend on some custom type.

I agree.

HeitorAugustoLN:
> Thinking about it I think we could make for example some attrset:
> 
> ```nix
> {
>   power = 0;
>   session = 1;
>   custom = 2;
>   powerAndSession = 3;
> }
> ```
> 
> and then set the enum values to the keys and then in the apply for the option just extract the value using this attrset. An alternative is to keep using `mkEnumOption`, but making it a proper option so that we don't have the need for `//` syntax. In that case though I would prefer it instead of taking in just a list of allowed values that it instead takes in an attrset, which makes it clearer the values assigned to the different enum values (and also allows not starting from `0`).
> 
> I think what's good about the first of these two is that we don't depend on our own custom `lib` throughout the project, which will make it easier to get into coming from `nixpkgs` or `home-manager`, but it will be less compact code.

I think it better using our current `getEnum` function for widgets that require enum values. Like this:

```nix
let
  getEnum = es: e:
      if e == null
      then null
      else
        lib.lists.findFirstIndex
          (x: x == e)
          (throw "getEnum: nonexistent key ${e}! This is a bug!")
          es;
in {}
```

Using this function, we can define options like this:

```nix
showSeconds = mkOption {
  type = with types; nullOr (enum [ "never" "onlyInTooltip" "always" ]);
  default = null;
  example = "always";
  description = ''
    When and where the seconds should be shown on the clock.
    Could be never, only in the tooltip on hover, or always.
  '';
  apply = getEnum [ "never" "onlyInTooltip" "always" ];
};
  ```

**Comment by HeitorAugustoLN** [comment]
> In that case though I would prefer it instead of taking in just a list of allowed values that it instead takes in an attrset,
> which makes it clearer the values assigned to the different enum values (and also allows not starting from 0).

For different cases, we can do the attrset, as we are already doing in `application-title-bar.nix` for example:

```nix
convertVerticalAlignment = verticalAlignment:
  let
    mappings = {
      "top" = "1";
      "center" = "128";
      "bottom" = "64";
      "baseline" = "256";
    };
  in
    mappings.${verticalAlignment} or (throw "Invalid enum value: ${verticalAlignment}");
```

magnouvean:
I'm happy with these proposals. With this we can even get rid of the `builtins.toString` (and the mappings in `covertVerticalAlignment` can become int types as well). I'll move it to `lib` though as there's nothing about this which is specific to widgets (it can be used for all sorts of other things as well if one wishes). It also wouldn't be much work needed starting from the current version of this pr.

It probably could do with a more descriptive name though, so I'll need to think about that.

HeitorAugustoLN:
> I'm happy with these proposals. With this we can even get rid of the `builtins.toString` (and the mappings in `covertVerticalAlignment` can become int types as well). I'll move it to `lib` though as there's nothing about this which is specific to widgets (it can be used for all sorts of other things as well if one wishes). It also wouldn't be much work needed starting from the current version of this pr.
> 
> It probably could do with a more descriptive name though, so I'll need to think about that.

Maybe `getIndexFromList` could be better.

magnouvean:
Looking back at it I might have misunderstood one of the comments above. Do you suggest we keep the `getEnum` (`getIndexFromList`) in `lib/tools.nix` for example, which will then be available for all modules to use, or just declared as a helping-function in the let-block in each module. I wouldn't mind the second approach of defining it in the let-block to be honest as I don't really want too much project-wide dependency on custom-made functions specific to `plasma-manager`, but I could be OK with including it as well (I don't have that strong of an opinion on it).

HeitorAugustoLN:
> Looking back at it I might have misunderstood one of the comments above. Do you suggest we keep the `getEnum` (`getIndexFromList`) in `lib/tools.nix` for example, which will then be available for all modules to use, or just declared as a helping-function in the let-block in each module. I wouldn't mind the second approach of defining it in the let-block to be honest as I don't really want too much project-wide dependency on custom-made functions specific to `plasma-manager`, but I could be OK with including it as well (I don't have that strong of an opinion on it).

Both approaches look good to me.

HeitorAugustoLN:
I tested it, not extensively. Seems to be working fine. Some options were not updated, so I created a [pull request](https://github.com/magnouvean/plasma-manager/pull/2) that updates it, if you want to merge it, or you can update it by yourself.

ada4a:
My (admittedly not too extensive) config seems to work perfectly. Even the generated files are identical.

modules/widgets/lib.nix
HeitorAugustoLN, line 1: Are values supposed to be converted to strings in the final file? If they are, it is not working, at least for me. But I don't think they should. As it the script works without any conversion when I tested.

magnouvean: No, I thought since there is support for bools, ints and floats on the desktop-script side that it's more appropriate to send them in as these types when possible.

HeitorAugustoLN: Yeah, I thought so. I just misunderstood the code.

magnouvean:
I think I've tested enough. There could still be a couple of bugs left that I've missed, but this should now be easy to fix. LGTM!

PR #250, opened 2024-07-07
Python in devshell

toast003:
We have python code on the repo, so I think it's a good idea to have python with the lsp in the devshell for people that don't have python installed system wide


toast003:
oh i guess i still don't know how to use git properly lol

magnouvean:
I reckon we can add `black` and `isort` as well since the script is formatted/refactored using both of these.

PR #256, opened 2024-07-10
Add rc2nix.py to replace rc2nix.rb

luochen1990:
Just a Python implementation, It is almost equivalent to the Ruby implementation (tested on my own device), but there are also some subtle differences:
1. For certain shortcut keys, the Ruby implementation returns [], while it returns content.
2. It fixed some minor [if condition bugs](https://github.com/nix-community/plasma-manager/pull/256/commits/24f7ab1434c0c952078f44fa4c1922edc01178c5) in the Ruby implementation.

And, Python Type Annotations are fully added, which will make life easier.


magnouvean:
Seems to work fine. It's long been planned to switch the `rc2nix` from ruby to python (ruby is a somewhat obscure language to use these days) so this is much appreciated. Seeing as the previous `rc2nix` was only just about working I think this is good enough to at least be as good as the previous implementation we had, so I'll merge right away :)

magnouvean:
Actually I have found one problem here with the escaping of `/`. If a `/` appears somewhere in a group in the config-file, for example if we want to write to the group `[Libinput][2][14][ETPS/2 Elantech Touchpad]` (in `kcminputrc`) as is the case in my config one would use the group `Libinput/2/14/ETPS\\/2 Elantech Touchpad` in the nix config (notice the two `\` before the `/`), but this `rc2nix` would generate `Libinput/2/14/ETPS\/2 Elantech Touchpad`, which would result in the group being `[Libinput][2][14][ETPS][2 Elantech Touchpad]`. I believe the previous `rc2nix` handled this correctly. Once this is fixed I'll merge right away.

luochen1990:
Perhaps we require a unit test mechanism. I will take care of this.

And, can you provide some test files for me (they need to be free of sensitive information and sufficiently comprehensive to cover typical scenarios)?

magnouvean:
Sure. I'll give some configs soon from my configuration. Just need a little bit of time to upload them to pastebin or something :)

magnouvean:
Found it out, [here](https://pastebin.com/S4NJbkCj) are some test files. See the comments for the filename. The config is perhaps too "clean", but I don't really use `rc2nix` that much so this was the best I could come up with.

luochen1990:
> The config is perhaps too "clean"

That will be a good start.  now I can reproduce this issue via unit test.

luochen1990:
@magnouvean Bug fixed. but the workflow requires approval from a maintainer.

magnouvean:
This looks good now. I will merge right away! Thanks for this :)

PR #260, opened 2024-07-12
Add titleReplacements option for application-title-bar

HeitorAugustoLN:
Made this few days ago, was just waiting for #248 to be merged into trunk.

Regexes have a few limitations, when using regular expressions that include tokens with a backslash (\), you'll need to escape the backslashes. This means using double backslashes (\\) in the script to ensure that the regular expression is correctly interpreted when the script runs.


magnouvean:
Hmm, adding the following widget:
```nix
      {
        applicationTitleBar = {
          titleReplacements = [
            {
              type = "string";
              originalTitle = "Brave Web Browser";
              newTitle = "Test";
            }
          ];
        };
      }
```
gives the following error:
```
error: function 'addStmt' called with unexpected argument 'TitleReplacements'
at /nix/store/kbvhf4krhvz5vi728jdw6kixn48nz7wq-source/modules/widgets/lib.nix:45:17:
```
Is it intentional that `TitleReplacements` is placed after `config`? That seems a bit strange.

HeitorAugustoLN:
Let me fix this

HeitorAugustoLN:
I rewrote it in the wrong place. I was using another branch for testing

HeitorAugustoLN:
And for the replacement to work, it will need to restart plasma shell after the script being run, at least back when I tested.

magnouvean:
Hmm we do have an interface for automatically restarting services after scripts now so it might be worth adding support for specifying this for widgets as well. It runs successfully now, and I have restarted the plasma shell now, but it doesn't seem to work for me (neither with the normal string of regexp). Is there a GUI where I can see if the rules are applied?

HeitorAugustoLN:
> Hmm we do have an interface for automatically restarting services after scripts now so it might be worth adding support for specifying this for widgets as well. It runs successfully now, and I have restarted the plasma shell now, but it doesn't seem to work for me (neither with the normal string of regexp). Is there a GUI where I can see if the rules are applied?

Yes, there is. Make sure you are using at least the version v0.6.5 of the widget. It got merged into master a few days ago, but i don't know if it is on unstable already

HeitorAugustoLN:
I am currently using an overlay to latest version, so it works for me.

magnouvean:
Ah I see. I'll trust you with that then. Are titleReplacements supported for `0.6.3` (which is what seems to be in `nixpkgs` a.t.m.)?

HeitorAugustoLN:
> Ah I see. I'll trust you with that then. Are titleReplacements supported for `0.6.3` (which is what seems to be in `nixpkgs` a.t.m.)?

titleReplacements were implemented in 0.6.5, so maybe we will need to wait a little bit.

But it will probably reach nixos-unstable soon

image

magnouvean:
I guess it makes sense then why it didn't work for me. I reckon we can merge when it gets into `nixos-unstable` (and then we can merge #262 right after. It probably will happen very soon since it was merged 3 days ago (usually it takes around 3 days in my experience).

HeitorAugustoLN:
@magnouvean 0.6.5 is on nixos-unstable right now

magnouvean:
Nice, it works great. Will merge right away :)

PR #262, opened 2024-07-12
Update home.nix example with widget-specific options and README

HeitorAugustoLN:


examples/home.nix
```diff
@@ -43,18 +43,17 @@
           # add the "icon" key to the "General" group for the widget in
           # ~/.config/plasma-org.kde.plasma.desktop-appletsrc.
           {
-            name = "org.kde.plasma.kickoff";
-            config = {
-              General.icon = "nix-snowflake-white";
+            kickoff = {
+              icon = "nix-snowflake-white";
+              sortAlphabetically = true;
```
magnouvean: I do agree that this is a better way of setting up the `kickoff` widget on, but I also think it's helpful to show the manual configuration syntax as well. We should maybe include the original `{name = ...; config = ...;}` and explain that it can also be done (perhaps better) using `{kickoff = {...};}`.

magnouvean (commented):
Looks good when the suggestion is solved :)

examples/home.nix
HeitorAugustoLN: Updated it.

magnouvean:
Could you also perhaps run `nixpkgs-fmt` on `home.nix`? Then it looks fine to me assuming the CI passes after #260 is merged.

PR #263, opened 2024-07-14
Refactor panels and add showLegend option for systemMonitor widget

### LINKED ISSUE #196: System Monitor add more options.
I'm missing `showTitle` and `showLegend` options, it would be nice to have them in systemMonitor wiget. 
It would be added to https://github.com/pjones/plasma-manager/blob/trunk/modules/widgets/system-monitor.nix file.

magnouvean:
Adds for panels a more extensible interface for adding custom widget packages by adding them into the `additionalWidgetPackages` attrset in `panels.nix`. Also adds the possibility of restarting certain services when certain widget configurations are present in some panel, namely by adding the appropriate `servicename = [{ "widget" = ...; cond = ...;  }]` in `serviceRestarts`, also in `panels.nix`. This way, if `cond` evaluates to `true` for any of the widgets with the name given in the widget attribute, the service this is placed under will be restarted after all the startup-scripts have run. This is a list so it is possible to add multiple tests.

I used this to restart `plasmashell` in the cases where `showLegend` is set in the systemmonitor widget, which for some reason is needed for it to correctly apply. As such this closes #196.

During refactoring, I also removed `programs.plasma.extraWidgets` as I didn't really see a use for it which couldn't be satisfied by just adding the widget packages to `home.packages` or `environment.systemPackages` manually. Imo the widgets can be automatically installed if the widget is in a panel, and I think that beyond this `extraWidgets` doesn't really make much sense (it only really makes sense if you plan on adding a widget which you configure via the GUI, but in that case I reckon installing it manually through nix is more than adequate). I don't know if there are other views on this though @HeitorAugustoLN.

### DIFF
--- a/examples/home.nix
+++ b/examples/home.nix
@@ -4,7 +4,6 @@
 
   programs.plasma = {
     enable = true;
-    extraWidgets = [ "application-title-bar" "plasmusic-toolbar" ];
 
     #
     # Some high-level settings:
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -7,6 +7,31 @@ let
   cfg = config.programs.plasma;
   hasWidget = widgetName: builtins.any (panel: builtins.any (widget: widget.name == widgetName) panel.widgets) cfg.panels;
 
+  # An attrset keeping track of the packages which should be added when a
+  # widget is present in the config.
+  additionalWidgetPackages = with pkgs; {
+    "com.github.antroids.application-title-bar" = [ application-title-bar ];
+    plasmusic-toolbar = [ plasmusic-toolbar ];
+  };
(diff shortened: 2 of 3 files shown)


HeitorAugustoLN:
> During refactoring, I also removed `programs.plasma.extraWidgets` as I didn't really see a use for it which couldn't be satisfied by just adding the widget packages to `home.packages` or `environment.systemPackages` manually. Imo the widgets can be automatically installed if the widget is in a panel, and I think that beyond this `extraWidgets` doesn't really make much sense (it only really makes sense if you plan on adding a widget which you configure via the GUI, but in that case I reckon installing it manually through nix is more than adequate). I don't know if there are other views on this though @HeitorAugustoLN.

Agreed. This is a better approach, I thought on removing this option myself.

PR #265, opened 2024-07-14
Add `borderlessMaximizedWindows` option to kwin module

HeitorAugustoLN:


modules/kwin.nix
```diff
@@ -307,5 +314,12 @@ in
           (virtualDesktopNameAttrs cfg.kwin.virtualDesktops.names)
         ];
       })
+
+      # Borderless maximized windows
+      (mkIf (cfg.kwin.borderlessMaximizedWindows != null) {
+        Windows = {
+          BorderlessMaximizedWindows = cfg.kwin.BorderlessMaximizedWindows;
```
magnouvean: Should be `cfg.kwin.borderlessMaximizedWindows` (small B).

magnouvean (commented):
Looks good apart from the suggestion given.

modules/kwin.nix
HeitorAugustoLN: You're right.

PR #267, opened 2024-07-19
Add custom font option for plasmusic-toolbar

HeitorAugustoLN:


modules/widgets/plasmusic-toolbar.nix
```diff
@@ -77,12 +270,26 @@ in
         displayInSeparateLines = mkBoolOption "Whether to display song information (title and artist) in separate lines or not.";
       };
       showPlaybackControls = mkBoolOption "Whether to show playback controls or not.";
+      font = {
+        useCustomFont = mkBoolOption "Whether to use custom font or not.";
```
magnouvean: Do we need this? We could maybe determine this automatically by checking if `font.settings` is `null` or not.

magnouvean:
Works fine on my side now. I'll merge when the check completes.

PR #268, opened 2024-07-20
Add settings option for all widget-specific options

HeitorAugustoLN:
Could be useful for users that want to configure options that are not covered by the widget-specific options, but don't want to configure the entire widget manually.

### DIFF
--- a/modules/widgets/application-title-bar.nix
+++ b/modules/widgets/application-title-bar.nix
@@ -437,7 +437,7 @@ in
           titleReplacementsTypes = map (r: r.type) replacements;
         };
       };
-      extraConfig = mkOption {
+      settings = mkOption {
         type = with types; nullOr (attrsOf (attrsOf (either (oneOf [ bool float int str ]) (listOf (oneOf [ bool float int str ])))));
         default = null;
         example = {
@@ -450,7 +450,7 @@ in
 
           See available options at https://github.com/antroids/application-title-bar/blob/main/package/contents/config/main.xml
         '';
-        apply = extraConfig: if extraConfig == null then {} else extraConfig;
+        apply = settings: if settings == null then {} else settings;
       };
     };
     convert =
@@ -463,7 +463,7 @@ in
       , mouseAreaClick
       , mouseAreaWheel
       , titleReplacements
-      , extraConfig
+      , settings
       }: {
         name = "com.github.antroids.application-title-bar";
         config = lib.recursiveUpdate {
@@ -537,7 +537,7 @@ in
             }
           );
           TitleReplacements = titleReplacements;
-        } extraConfig;
+        } settings;
       };
   };
 }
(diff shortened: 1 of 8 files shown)


HeitorAugustoLN:
I just now saw that I didn't put the right example for application-title-bar

magnouvean (commented):
I see that the use of `//` for adding the `extraConfig` configuration to the widget causes some problems. For example:
```nix
        plasmusicToolbar = {
          preferredSource = "any";
          panelIcon.albumCover = {
            useAsIcon = true;
            radius = 2;
          };
          showPlaybackControls = false;
          extraConfig = {
            General.somesetting = 4;
          };
        };
```
Will result in only `somesetting` key is set in the`General` group for the widget configuration (so most of the configuration is simply ignored). This is due to for example:
```nix
{
  General = { key1 = 1; key2 = 2; }
} // { General.key3 = 3; }
```
Will evaluate to `{General.key3 = 3;}` since `General` exists in both the attrsets and thus only the latter will be used.
I believe `lib.mkMerge` will fix this, but it would be less pretty code (but maybe that's just needed here).

ada4a:
@magnouvean
> `lib.mkMerge`

wouldn't [`lib.recursiveUpdate`](https://ryantm.github.io/nixpkgs/functions/library/attrsets/#function-library-lib.attrsets.recursiveUpdate) be more suitable here?

@HeitorAugustoLN 
this PR seems similar to the pattern that [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md) proposes. Do you think it would make sense to bring it closer to that by using `settings` instead of `extraConfig`?

magnouvean:
Yeah I agree with both of the above suggestions (`settings` and `recursiveUpdate`)

HeitorAugustoLN:
Can anybody test the systemmonitor with lib.recursiveUpdate? I don't know if I did it the right way.

magnouvean:
I believe the semicolon should still be there, but you were missing a `)`. We'll see with the CI though

HeitorAugustoLN:
Yeah, i noticed it now.

magnouvean:
The `systemMonitor` widget now seems to work fine, though there still are inifiniteRecursion for the others. I tested a panel with only `systemMonitor` and it works as expected.

HeitorAugustoLN:
plasmusic seems fine too

HeitorAugustoLN:
It seems like the problem is system-tray

HeitorAugustoLN:
I will fix it.

HeitorAugustoLN:
Apparently now it works.

magnouvean:
Seems to work well now, and the code looks good as well. Thanks for this and the other PRs :)

HeitorAugustoLN:
Happy to help :)

PR #269, opened 2024-07-20
Add editor.font option for Kate

HeitorAugustoLN:


Asqiir:
Nice work!
This is one of the missing things of #87 that I never figured out

Issue #192, opened 2024-06-06
Searchability for plasma-manager options


Asqiir:
I'd like to have better searchability of our config options.
How hard would it be to set up something like [the nixos options search](https://search.nixos.org/options) or get added to [mynixos search](https://mynixos.com/)?
(Is that even in-scope?)

magnouvean:
I would say it is in scope, though at least nixos options search is probably more work than at least I have time for. I haven't heard about `mynixos` before, but maybe if it's not too difficult to set up (I'm assuming it is quite easy) we could maybe get it up there. I'll certainly check that out

Asqiir:
I looked up the official nixos search. [It's this project.](https://github.com/NixOS/nixos-search)

The readme has a section on [adding own flakes](https://github.com/NixOS/nixos-search#adding-flakes), although I cannot decipher whether that means "to the official setup" or "in your own installation".

If it's the first – and if (a big if) plasma-manager is in scope for them – that might be our solution.

magnouvean:
Looking at how it works it looks to be the the first indeed. We can see if they are open to adding it there (the only objection I can see is that `plasma-manager` probably has a lot more options than the other flakes there currently). Even if that wouldn't work we could maybe try to add it in a fork, but that is dependent on it being possible to host on `github.io`. If none of those options are available we'll probably have to do with just `mynixos`.

ede1998:
I'd love to see that feature.

Another idea: Home-Manager generates a [large html page](https://nix-community.github.io/home-manager/options.xhtml) with all the available options. This would still be searchable with the browser built-in search.

In the code that does not look too complex and should be easily hostable with Github pages:
https://github.com/nix-community/home-manager/blob/master/docs/default.nix#L57 ([permalink](https://github.com/nix-community/home-manager/blob/635563f245309ef5320f80c7ebcb89b2398d2949/docs/default.nix#L57))

Sadly, I'm not well versed in nix or I'd give it a try myself.

Asqiir:
One thing I learned only recently: how incredibly good the github search in-repo search is (press / and then search for some option). This apparently means that lots of devs don't have the need for nixos-search and thus there isn't a whole lot of pressure for its implementation

magnouvean:
[...]

PR #278, opened 2024-07-28
Add plasma-manager documentation generation

magnouvean:
The documentation for this project is known for being somewhat lacking. This aims to fix this by adding a documentation/manual which is automatically built on each commit to `trunk`. This manual includes an options-page which lists all the available options supported in `plasma-manager`, as mentioned in #192.

A working preview can a.t.m. be found at https://magnouvean.github.io/plasma-manager (this will probably be deleted after this pr is closed). Feedback for this would be great. I reckon if there are no major complaints that I merge this in a couple days. I'm not 100% sure the action will work on this repo since this is under `nix-community` (it's hard to know beforehand, but the action is pretty much identical to the one in `home-manager`, so hopefully there's no problem there).


HeitorAugustoLN:
Great work! It is looking great.

HeitorAugustoLN:
In the future we could probably add this for widget options too.

magnouvean:
Yes, though we'd probably need to quite creative with a solution for this since `lib.evalModules` wouldn't work in that case (since widget options aren't modules). We could probably get something working though eventually.

PR #280, opened 2024-07-28
Fix window module being missing

HeitorAugustoLN:
Was rewriting my nixos configuration and I noted that this module was missing.


magnouvean:
Huh, I guess since it contains so few options it's gone under the radar. Nice spot.

PR #285, opened 2024-07-29
okular: add new options

KucharczykL:
I have added options for these Okular settings:

mouseMode
zoomMode
viewMode
obeyDrm

This is my first Nix contribution so please let me know if I should change anything.

### DIFF
--- a/modules/apps/okular.nix
+++ b/modules/apps/okular.nix
@@ -1,29 +1,40 @@
-{ config, lib, pkgs, ... }:
+{ config
+, lib
+, pkgs
+, ...
+}:
 
 let
   cfg = config.programs.okular;
-  getIndexFromEnum = enum: value:
-  if value == null
-  then null
-  else
-    lib.lists.findFirstIndex
-      (x: x == value)
-      (throw "getIndexFromEnum (okular): Value ${value} isn't present in the enum. This is a bug.")
-      enum;
-in with lib.types;
+  getIndexFromEnum =
+    enum: value:
+    if value == null then
+      null
+    else
+      lib.lists.findFirstIndex (x: x == value)
+        (throw "getIndexFromEnum (okular): Value ${value} isn't present in the enum. This is a bug.")
+        enum;
+in
+with lib.types;
 {
   options.programs.okular = {
     enable = lib.mkEnableOption ''
       Enable configuration management for okular.
     '';
 
-    package = lib.mkPackageOption pkgs [ "kdePackages" "okular" ] {
-      example = "pkgs.libsForQt5.okular";
-      extraDescription = ''
-        Which okular package to install. Use `pkgs.libsForQt5.okular` in Plasma5 and
-        `pkgs.kdePackages.okular` in Plasma6.
-      '';
-    };
+    package =
+      lib.mkPackageOption pkgs
+        [
+          "kdePackages"
+          "okular"
+        ]
+        {
+          example = "pkgs.libsForQt5.okular";
+          extraDescription = ''
+            Which okular package to install. Use `pkgs.libsForQt5.okular` in Plasma5 and
+            `pkgs.kdePackages.okular` in Plasma6.
+          '';
+        };
 
     # ==================================
     #     GENERAL
@@ -55,21 +66,32 @@ in with lib.types;
       viewMode = lib.mkOption {
         description = "The view mode for the pages.";
         default = null;
-        type =
-          nullOr (enum [ "Single" "Facing" "FacingFirstCentered" "Summary" ]);
+        type = nullOr (enum [
+          "Single"
+          "Facing"
+          "FacingFirstCentered"
+          "Summary"
+        ]);
       };
 
-      zoomMode = 
-      let enumVals = [ "100%" "fitWidth" "fitPage" "autoFit" ];
-      in lib.mkOption {
-        description = ''
-          Specifies the default zoom mode for file which were never opened before.
-          For those files which were opened before the previous zoom mode is applied.
+      zoomMode =
+        let
+          enumVals = [
+            "100%"
+            "fitWidth"
+            "fitPage"
+            "autoFit"
+          ];
+        in
+        lib.mkOption {
+          description = ''
+            Specifies the default zoom mode for file which were never opened before.
+            For those files which were opened before the previous zoom mode is applied.
           '';
-        default = null;
-        type = nullOr (enum enumVals);
-        apply = getIndexFromEnum enumVals;
-      };
+          default = null;
+          type = nullOr (enum enumVals);
+          apply = getIndexFromEnum enumVals;
+        };
 
       obeyDrm = lib.mkOption {
         description = ''
@@ -119,7 +141,11 @@ in with lib.types;
         mode = lib.mkOption {
           description = "Mode used to change the colors.";
           default = null;
-          type = nullOr (enum [ "Inverted" "Paper" "Recolor" ]);
+          type = nullOr (enum [
+            "Inverted"
+            "Paper"
+            "Recolor"
+          ]);
         };
         paperColor = lib.mkOption {
           description = "Paper color in RGB. Used for the `Paper` mode.";
@@ -154,7 +180,12 @@ in with lib.types;
       memoryUsage = lib.mkOption {
         description = "Memory usage of Okular. This impacts the speed performance of Okular as it determines how much computation results are kept in memory and not recomputed.";
         default = null;
-        type = nullOr (enum [ "Low" "Normal" "Agressive" "Greedy" ]);
+        type = nullOr (enum [
+          "Low"
+          "Normal"
+          "Agressive"
+          "Greedy"
+        ]);
       };
     };
   };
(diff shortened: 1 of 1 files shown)


modules/apps/okular.nix
```diff
@@ -43,6 +43,53 @@ in with lib.types;
...
+        description = ''
+          Specifies the default zoom mode for file which were never opened before. For those files which were opened before the previous zoom mode is applied.
+          0: "100%"
+          1: "Fit Width"
+          2: "Fit Page"
+          3: "Auto Fit"
+          '';
+        default = 1;
+        type = nullOr (enum [ 0 1 2 3 ]);
+      };
```
HeitorAugustoLN: Default should be set to `null`
Use getIndexFromEnum function for zoomMode:
```suggestion
      zoomMode =
        let enumVals = [ "100%" "fitWidth" "fitPage" "autoFit" ];
        in lib.mkOption {
          description = ''
            Specifies the default zoom mode for file which were never opened before. For those files which were opened before the previous zoom mode is applied.
            '';
          default = null;
          type = nullOr (enum enumVals);
          apply = getIndexFromEnum enumVals;
        };
```

Add this function in the let statement at the start of the file.
```nix
getIndexFromEnum = enum: value:
  if value == null
  then null
  else
    lib.lists.findFirstIndex
      (x: x == value)
      (throw "getIndexFromEnum (okular): Value ${value} isn't present in the enum. This is a bug")
      enum;
```

modules/apps/okular.nix
```diff
@@ -43,6 +43,53 @@ in with lib.types;
         default = null;
         type = nullOr bool;
       };
+
+      viewMode = lib.mkOption {
+        description = "The view mode for the pages.";
+        default = "Single";
+        type =
+          nullOr (enum [ "Single" "Facing" "FacingFirstCentered" "Summary" ]);
+      };
```
HeitorAugustoLN:
```suggestion
      viewMode = lib.mkOption {
        description = "The view mode for the pages.";
        default = null;
        type =
          nullOr (enum [ "Single" "Facing" "FacingFirstCentered" "Summary" ]);
      };
```

modules/apps/okular.nix
```diff
@@ -43,6 +43,53 @@ in with lib.types;
...
+          Text Selection - The mouse will work as a text selection tool. In that mode clicking left mouse button and dragging will give the option of selecting the text of the document. Then, just click with the right mouse button to copy to the clipboard or speak the current selection.
+          Table Selection - Draw a rectangle around the text for the table, then click with the left mouse button to divide the text block into rows and columns. A left mouse button click on an existing line removes it and merges the adjacent rows or columns. Finally, just click with the right mouse button to copy the table to the clipboard.
+          Magnifier - Activates the magnifier mode for the mouse pointer. Press and hold the left mouse button to activate magnifier widget, move the pointer for panning through the document. The magnifier scales each pixel in the document into 10 pixels in the magnifier widget.
+        '';
+        default = "Browse";
```
HeitorAugustoLN:
```suggestion
        default = null;
```

HeitorAugustoLN:
Everything else seems fine. Great work!

modules/apps/okular.nix
```diff
@@ -43,6 +43,53 @@ in with lib.types;
...
+          3: "Auto Fit"
+          '';
+        default = 1;
+        type = nullOr (enum [ 0 1 2 3 ]);
+      };
+
+      obeyDrm = lib.mkOption {
+        description =
+          "Whether Okular should obey DRM (Digital Rights Management) restrictions. DRM limitations are used to make it impossible to perform certain actions with PDF documents, such as copying content to the clipboard. Note that in some configurations of Okular, this option is not available.";
+        default = true;
```
HeitorAugustoLN:
```suggestion
        default = null;
```

Almost forgot that one too.

modules/apps/okular.nix
```diff
@@ -43,6 +43,53 @@ in with lib.types;
...
+      viewMode = lib.mkOption {
+        description = "The view mode for the pages.";
+        default = "Single";
+        type =
+          nullOr (enum [ "Single" "Facing" "FacingFirstCentered" "Summary" ]);
+      };
+
+      zoomMode = lib.mkOption {
+        description = ''
+          Specifies the default zoom mode for file which were never opened before. For those files which were opened before the previous zoom mode is applied.
```
magnouvean: Maybe split this first line into two so that the line doesn't become too long

modules/apps/okular.nix
```diff
@@ -43,6 +43,53 @@ in with lib.types;
...
+          2: "Fit Page"
+          3: "Auto Fit"
+          '';
+        default = 1;
+        type = nullOr (enum [ 0 1 2 3 ]);
+      };
+
+      obeyDrm = lib.mkOption {
+        description =
+          "Whether Okular should obey DRM (Digital Rights Management) restrictions. DRM limitations are used to make it impossible to perform certain actions with PDF documents, such as copying content to the clipboard. Note that in some configurations of Okular, this option is not available.";
```
magnouvean: Move to a multi-line description and split the lines. I.e. something like:
```nix
description = ''
  Whether Okular should obey DRM (Digital Rights Management) restrictions.
  DRM limitations are used to make it impossible to perform certain actions
  with PDF documents, such as copying content to the clipboard. Note that
  in some configurations of Okular, this option is not available.
'';
```

modules/apps/okular.nix
```diff
@@ -43,6 +43,53 @@ in with lib.types;
...
+          Text Selection - The mouse will work as a text selection tool. In that mode clicking left mouse button and dragging will give the option of selecting the text of the document. Then, just click with the right mouse button to copy to the clipboard or speak the current selection.
+          Table Selection - Draw a rectangle around the text for the table, then click with the left mouse button to divide the text block into rows and columns. A left mouse button click on an existing line removes it and merges the adjacent rows or columns. Finally, just click with the right mouse button to copy the table to the clipboard.
+          Magnifier - Activates the magnifier mode for the mouse pointer. Press and hold the left mouse button to activate magnifier widget, move the pointer for panning through the document. The magnifier scales each pixel in the document into 10 pixels in the magnifier widget.
```
magnouvean: I think this description maybe is more than what's neccesary. Imo it's enough to just have the first sentence for each item, so for example the first second one would be:
```
Zoom - The mouse will work as a zoom tool.
```
(usually option descriptions are more on the compact side)

I also think a sentence at the start of the option would be nice to comment on what mouseMode itself sets, not just what the options do. Then you can end this with `This allows the following options:`

modules/apps/okular.nix
```diff
@@ -43,6 +43,53 @@ in with lib.types;
...
+        '';
+        default = "Browse";
+        type = nullOr (enum [
+          "Browse"
+          "Zoom"
+          "RectSelect"
+          "TextSelect"
+          "TableSelect"
+          "Magnifier"
+          "TrimSelect"
```
magnouvean, line 126: Trimselect doesn't seem to be included in the description above. Also it appear as `RectSelect` is called Area Select in the description above. I reckon calling it `Rectangle Selection` in the description above is more appropriate.

magnouvean:
From my testing it seems like the functionality is there so when the above things are fixed we probably can merge rather quickly.

modules/apps/okular.nix
KucharczykL: I took it from the official docs but I think it's better to make it consistent with the config seen here so I've renamed it.

modules/apps/okular.nix
KucharczykL: I've rewritten the description to be brief and added a link to the official docs, hope that's OK!

modules/apps/okular.nix
KucharczykL: Split over several lines.

modules/apps/okular.nix
KucharczykL: Split the sentence into two lines.

modules/apps/okular.nix
KucharczykL: Thank you! I want to do something like this but didn't know how to go about it. I was hoping that there is some kind of `enum` that allows setting both a value and display string.

modules/apps/okular.nix
KucharczykL: Changed to null.

modules/apps/okular.nix
KucharczykL: Changed to null.

modules/apps/okular.nix
KucharczykL: Changed to null.

KucharczykL:
Question: the file does not seem formatted. Should I format it with `nixfmt-rfc-style`?

modules/apps/okular.nix
HeitorAugustoLN: You could use the `oneOf` function for that. For example:

```nix
type = nullOr (oneOf (enum enumVals) (enum [0 1 2 3]));
```

But you would need to modify the apply function to be like this:

```nix
apply = zoomMode: if builtins.isString zoomMode then getIndexFromEnum enumVals zoomMode else zoomMode
```

magnouvean:
We use `nixpkgs-fmt` with the default settings for this project. I would be nice if you could format using this as well (if it isn't already) :)

modules/apps/okular.nix
```diff
@@ -46,48 +54,53 @@ in with lib.types;
...
-        default = "Browse";
+        default = null;
         type = nullOr (enum [
           "Browse"
           "Zoom"
           "RectSelect"
           "TextSelect"
           "TableSelect"
           "Magnifier"
-          "TrimSelect"
```
HeitorAugustoLN: I think this should not be removed as it is an actual [option](https://github.com/KDE/okular/blob/master/conf/okular.kcfg#L365).

KucharczykL: Yeah, that's why I originally included it but it's not documented anywhere AFAICT so I'm not sure what it does. I will add it back.

magnouvean: I believe the setting for this is under `Tools` in the top menu, where on my version at the very least it's not available. It might be something which is present in a future version though so it's fine to keep it in. We can add a line for what the option does when the documentation is updated or we find out exactly what it does.

magnouvean:
All looks good now! I will merge right away. Thanks for the contribution :)

PR #291, opened 2024-07-31
Add battery options for powerdevil module

### LINKED ISSUE #288: powerdevil: also allow setting options for BAT
right now settings for powerdevil only affect AC related settings but not for BAT.

For example setting
```nix
powerdevil = {
    autoSuspend = {
        action = "nothing";
        idleTimeout = null;
    };
    turnOffDisplay.idleTimeout = 900;
};
```

results in 
```bash
[AC][Display]
TurnOffDisplayIdleTimeoutSec=900

[AC][SuspendAndShutdown]
AutoSuspendAction=0
```

I would also like to set the settings for BAT.

magnouvean:
Closes #288. I tried to make this quite scalable, though perhaps somewhat at the cost of readability.

Keep in mind I do not have a laptop to test this on so I will need someone to check whether the `powerdevil.battery` options work (preferably `powerButtonAction`, `turnOffDisplay` and `autoSuspend`). For example, here is a configuration utilizing all three options:
```nix
programs.plasma.powerdevil = {
  battery = {
    powerButtonAction = "sleep";
    turnOffDisplay = {
      idleTimeout = 600;
      idleTimeoutWhenLocked = "whenLockedAndUnlocked";
    };
    autoSuspend = {
      action = "shutDown";
      idleTimeout = 300;
    };
  };
};
```
@Mag1cByt3s

### DIFF
--- a/modules/powerdevil.nix
+++ b/modules/powerdevil.nix
@@ -16,93 +16,117 @@ let
     sleep = null;
     shutDown = 8;
   };
-in
-{
-  config.assertions = [
-    {
-      assertion = (cfg.powerdevil.autoSuspend.action != autoSuspendActions.nothing || cfg.powerdevil.autoSuspend.idleTimeout == null);
-      message = "Setting programs.plasma.powerdevil.autoSuspend.idleTimeout for autosuspend-action \"nothing\" is not supported.";
-    }
-    {
-      assertion = (cfg.powerdevil.turnOffDisplay.idleTimeout != -1 || cfg.powerdevil.turnOffDisplay.idleTimeoutWhenLocked == null);
-      message = "Setting programs.plasma.powerdevil.turnOffDisplay.idleTimeoutWhenLocked for when idleTimeout is \"never\" is not supported.";
-    }
-  ];
 
-  options = {
-    programs.plasma.powerdevil = {
-      powerButtonAction = lib.mkOption {
-        type = with lib.types; nullOr (enum (builtins.attrNames powerButtonActions));
+  # Since AC and battery allows the same options we create a function here which
+  # can generate the options by just specifying the type (i.e. "AC" or
+  # "battery").
+  createPowerDevilOptions = type: {
+    powerButtonAction = lib.mkOption {
+      type = with lib.types; nullOr (enum (builtins.attrNames powerButtonActions));
+      default = null;
+      example = "nothing";
+      description = ''
+        The action, when on ${type}, to perform when the power button is pressed.
+      '';
+      apply = action: if (action == null) then null else powerButtonActions."${action}";
+    };
+    autoSuspend = {
+      action = lib.mkOption {
+        type = with lib.types; nullOr (enum (builtins.attrNames autoSuspendActions));
         default = null;
         example = "nothing";
         description = ''
-          The action to perform when the power button is pressed.
+          The action, when on ${type}, to perform after a certain period of inactivity.
         '';
-        apply = action: if (action == null) then null else powerButtonActions."${action}";
+        apply = action: if (action == null) then null else autoSuspendActions."${action}";
       };
-      autoSuspend = {
-        action = lib.mkOption {
-          type = with lib.types; nullOr (enum (builtins.attrNames autoSuspendActions));
-          default = null;
-          example = "nothing";
-          description = ''
-            The action to perform after a certain period of inactivity.
-          '';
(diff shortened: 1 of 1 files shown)


Mag1cByt3s:
I just tested it like so:
```nix
powerdevil = {
      battery = {
        autoSuspend = {
          action = "nothing";
          idleTimeout = null;
        };
        turnOffDisplay.idleTimeout = 900;
      };

      AC = {
        autoSuspend = {
          action = "nothing";
          idleTimeout = null;
        };
        turnOffDisplay.idleTimeout = 900;
      };
    };
```

that results in .config/powerdevilrc:
```bash
[AC][Display]
TurnOffDisplayIdleTimeoutSec=900

[AC][SuspendAndShutdown]
AutoSuspendAction=0

[BAT][Display]
TurnOffDisplayIdleTimeoutSec=900

[BAT][SuspendAndShutdown]
AutoSuspendAction=0
```

So it seems to work and fix the issue.

magnouvean:
If it's possible checking, does the settings apply correctly when logging in/out? I made in this pr with the assumption that the keys (and available settings) were the same for AC and battery. If it seems to work I can merge right away I think.

Mag1cByt3s:
I just checked. The config file `~/.config/powerdevilrc` stays the same after logging out. The settings made there are not being reflected by KDE Power Management settings. I am not sure if that is intentional or not.

magnouvean:
If you for example reboot and login normally, are the settings in the settings app what you expect from the configuration? If it isn't, could you try to change the settings through the settings app and see what kinds of keys/values is written to `powerdevilrc`. It could be that this pr writes the wrong keys.

Mag1cByt3s:
I just rebooted but the effect is the same. The settings are present in the config file but the settings applications still does not reflect these values.

magnouvean:
Hmm that indicates the keys we write are wrong. Could you try changing them through the settings app and report the new keys which is written?

Mag1cByt3s:
Ok so here is the powerdevilrc before changing the settings:

```bash
[AC][Display]
TurnOffDisplayIdleTimeoutSec=900

[AC][SuspendAndShutdown]
AutoSuspendAction=0

[BAT][Display]
TurnOffDisplayIdleTimeoutSec=900

[BAT][SuspendAndShutdown]
AutoSuspendAction=0
```

and here is the config after disabling sleep on inactivity for both battery and AC:

```bash
[AC][Display]
TurnOffDisplayIdleTimeoutSec=900

[AC][SuspendAndShutdown]
AutoSuspendAction=0

[BAT][Display]
TurnOffDisplayIdleTimeoutSec=900

[BAT][SuspendAndShutdown]
AutoSuspendAction=0

[Battery][Display]
DimDisplayIdleTimeoutSec=300
TurnOffDisplayIdleTimeoutSec=900

[Battery][SuspendAndShutdown]
AutoSuspendAction=0
```

It seems like "BAT" should be "Battery"

magnouvean:
I see, it's just that the section isn't called `BAT`, but `Battery`. It also seems like there may be some screen-dimming settings as well which we may want to add options for as well.

Mag1cByt3s:
Yes. I also set some screen dimming options. I am not sure if these were present or not already in plasma-manager. If not it would be good to add them.

Edit: `TurnOffDisplayIdleTimeoutSec` was already present. `DimDisplayIdleTimeoutSec` not.

magnouvean:
I'll try to make something work for screen dimming.

magnouvean:
All right I've added `dimScreen` options for both battery and AC and fixed the keys being written to the wrong group. Again I haven't been able to test the functionality though. If they (and the rest of the battery configuration) work as expected I think I'll merge right away, but if not (and there isn't any clear error I've made), maybe I'll try to find a way to get this options-menu on my desktop (through a VM or something) and then do the full testing (this probably would take some time though and I'll be quite busy from next week onwards so hopefully we can figure it out).

Mag1cByt3s:
I will try to test it. But i can only really test it commit by commit.

Mag1cByt3s:
I just tried it out.

It seems to work without issues.

I just checked the settings now also match the settings in KDE power settings.

magnouvean:
Very nice, I'll merge this right now then and if someone encounters some issues we can fix them in a separate pr. Thanks for testing!

Mag1cByt3s:
I just tested it again now that it is merged.

The results still look good.

PR #292, opened 2024-07-31
Add a few desktop folder settings

Morwdan:
This PR adds four options from the `Icons` menu of the `Desktop Folder Settings`. They seem to work (for me at least), but I’m not sure if the implementation is good enough for inclusion here. I’m specifically unsure about the following:

- Is `workspace.nix` the right place for the new options?
- Is it okay to use the `desktop` set?
- Is the implementation via the panel script acceptable?

I’d also like to add a few options for the `Mouse Actions` menu to the `desktop` set, but I suspect [my approach](https://github.com/Morwdan/plasma-manager/commit/2703e2c4709eb2841fdb0ebd755688d65d8a603c) might be a bit to hacky and might also not work in other setups.

### DIFF
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -161,7 +161,8 @@ let
   anyDesktopFolderSettingsSet = (
     (cfg.workspace.desktop.icons.arrangement != null) ||
     (cfg.workspace.desktop.icons.size != null) ||
-    (cfg.workspace.desktop.icons.folderPreviewPopups != null)
+    (cfg.workspace.desktop.icons.folderPreviewPopups != null) ||
+    (cfg.workspace.desktop.icons.previewPlugins != null)
   );
 in
 {
@@ -248,6 +249,7 @@ in
                 ${lib.optionalString (cfg.workspace.desktop.icons.arrangement == "topToBottom") ''desktop.writeConfig("arrangement", 1);''}
                 ${stringIfNotNull cfg.workspace.desktop.icons.size ''desktop.writeConfig("iconSize", ${builtins.toString cfg.workspace.desktop.icons.size});''}
                 ${lib.optionalString (!cfg.workspace.desktop.icons.folderPreviewPopups) ''desktop.writeConfig("popups", false);''}
+                ${stringIfNotNull cfg.workspace.desktop.icons.previewPlugins ''desktop.writeConfig("previewPlugins", "${lib.strings.concatStringsSep "," cfg.workspace.desktop.icons.previewPlugins}");''}
               }
           '' else ""
           );
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -7,8 +7,6 @@ let
   cfg = config.programs.plasma;
   hasWidget = widgetName: builtins.any (panel: builtins.any (widget: widget.name == widgetName) panel.widgets) cfg.panels;
 
-  inherit (widgets.lib) stringIfNotNull;
-
   # An attrset keeping track of the packages which should be added when a
   # widget is present in the config.
   additionalWidgetPackages = with pkgs; {
@@ -157,9 +155,6 @@ let
     (cfg.workspace.wallpaperPictureOfTheDay != null) ||
     (cfg.workspace.wallpaperPlainColor != null) ||
     ((builtins.length cfg.panels) > 0));
-
-  # Becomes true if any option under "cfg.workspace.desktop.icons" is set to something other than null.
-  anyDesktopFolderSettingsSet = lib.any (v: v != null) (builtins.attrValues cfg.workspace.desktop.icons);
 in
 {
   imports = [
@@ -236,25 +231,12 @@ in
                 desktop.currentConfigGroup = Array("Wallpaper", "org.kde.color", "General");
                 desktop.writeConfig("Color", "${cfg.workspace.wallpaperPlainColor}");
             }
-          '' else "");
-          desktopFolderSettings = (if anyDesktopFolderSettingsSet then ''
-            // Desktop folder settings
-            let allDesktops = desktops();
-            for (const desktop of allDesktops) {
-                desktop.currentConfigGroup = ["General"];
-                ${lib.optionalString (cfg.workspace.desktop.icons.arrangement == "topToBottom") ''desktop.writeConfig("arrangement", 1);''}
-                ${lib.optionalString (cfg.workspace.desktop.icons.alignment == "right") ''desktop.writeConfig("alignment", 1);''}
-                ${lib.optionalString (cfg.workspace.desktop.icons.lockInPlace) ''desktop.writeConfig("locked", true);''}
-                ${stringIfNotNull cfg.workspace.desktop.icons.size ''desktop.writeConfig("iconSize", ${builtins.toString cfg.workspace.desktop.icons.size});''}
-                ${lib.optionalString (!cfg.workspace.desktop.icons.folderPreviewPopups) ''desktop.writeConfig("popups", false);''}
-                ${stringIfNotNull cfg.workspace.desktop.icons.previewPlugins ''desktop.writeConfig("previewPlugins", "${lib.strings.concatStringsSep "," cfg.workspace.desktop.icons.previewPlugins}");''}
-              }
           '' else ""
           );
         in
         {
           preCommands = panelPreCMD;
-          text = panelLayoutStr + wallpaperDesktopScript + wallpaperSlideShow + wallpaperPOTD + wallpaperPlainColor + desktopFolderSettings;
+          text = panelLayoutStr + wallpaperDesktopScript + wallpaperSlideShow + wallpaperPOTD + wallpaperPlainColor;
           postCommands = panelPostCMD + wallpaperPostCMD;
           restartServices =
             (lib.unique (if anyNonDefaultScreens then [ "plasma-plasmashell" ] else [ ])
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -155,6 +155,10 @@ let
     (cfg.workspace.wallpaperPictureOfTheDay != null) ||
     (cfg.workspace.wallpaperPlainColor != null) ||
     ((builtins.length cfg.panels) > 0));
+
+  anyDesktopFolderSettingsSet = (
+    (cfg.workspace.desktop.icons.arrangement != null)
+  );
 in
 {
   imports = [
@@ -231,12 +235,20 @@ in
                 desktop.currentConfigGroup = Array("Wallpaper", "org.kde.color", "General");
                 desktop.writeConfig("Color", "${cfg.workspace.wallpaperPlainColor}");
             }
+          '' else "");
+          desktopFolderSettings = (if anyDesktopFolderSettingsSet then ''
+            // Desktop folder settings
+            let allDesktops = desktops();
+            for (const desktop of allDesktops) {
+                desktop.currentConfigGroup = ["General"];
+                ${lib.optionalString (cfg.workspace.desktop.icons.arrangement == "topToBottom") ''desktop.writeConfig("arrangement", 1);''}
+              }
           '' else ""
           );
         in
         {
           preCommands = panelPreCMD;
-          text = panelLayoutStr + wallpaperDesktopScript + wallpaperSlideShow + wallpaperPOTD + wallpaperPlainColor;
+          text = panelLayoutStr + wallpaperDesktopScript + wallpaperSlideShow + wallpaperPOTD + wallpaperPlainColor + desktopFolderSettings;
           postCommands = panelPostCMD + wallpaperPostCMD;
           restartServices =
             (lib.unique (if anyNonDefaultScreens then [ "plasma-plasmashell" ] else [ ])
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -160,7 +160,8 @@ let
 
   anyDesktopFolderSettingsSet = (
     (cfg.workspace.desktop.icons.arrangement != null) ||
-    (cfg.workspace.desktop.icons.size != null)
+    (cfg.workspace.desktop.icons.size != null) ||
+    (cfg.workspace.desktop.icons.folderPreviewPopups != null)
   );
 in
 {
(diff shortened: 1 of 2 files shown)


modules/panels.nix
```diff
@@ -155,6 +157,13 @@ let
     (cfg.workspace.wallpaperPictureOfTheDay != null) ||
     (cfg.workspace.wallpaperPlainColor != null) ||
     ((builtins.length cfg.panels) > 0));
+
+  anyDesktopFolderSettingsSet = (
+    (cfg.workspace.desktop.icons.arrangement != null) ||
+    (cfg.workspace.desktop.icons.size != null) ||
+    (cfg.workspace.desktop.icons.folderPreviewPopups != null) ||
+    (cfg.workspace.desktop.icons.previewPlugins != null)
+  );
```
magnouvean: Here I believe `anyDesktopFolderSettingsSet = lib.any (v: v != null) (builtins.attrValues cfg.workspace.desktop.icons)` would be more elegant.

modules/panels.nix
```diff
@@ -231,12 +240,23 @@ in
...
+            let allDesktops = desktops();
+            for (const desktop of allDesktops) {
+                desktop.currentConfigGroup = ["General"];
+                ${lib.optionalString (cfg.workspace.desktop.icons.arrangement == "topToBottom") ''desktop.writeConfig("arrangement", 1);''}
+                ${stringIfNotNull cfg.workspace.desktop.icons.size ''desktop.writeConfig("iconSize", ${builtins.toString cfg.workspace.desktop.icons.size});''}
+                ${lib.optionalString (!cfg.workspace.desktop.icons.folderPreviewPopups) ''desktop.writeConfig("popups", false);''}
+                ${stringIfNotNull cfg.workspace.desktop.icons.previewPlugins ''desktop.writeConfig("previewPlugins", "${lib.strings.concatStringsSep "," cfg.workspace.desktop.icons.previewPlugins}");''}
+              }
           '' else ""
           );
```
magnouvean: I think it would be better to move this code into a separate script. I think you could instead put this into something like:
```nix
programs.plasma.startup.desktopScript."set_desktop_folder_settings" = (lib.mkIf anyDesktopFolderSettingsSet {
  text = ''
    {{SCRIPT_CONTENT_HERE}}
  '';
  priority = 3;
});
```
Where you would also move `anyDesktopFolderSettingsSet` to the `workspace.nix` of course.

The reason why the wallpaper and panels had to be put in the same file was that when they were in separate scripts and the panel was updated, the wallpaper script also had to be run for the wallpaper to be set (when `overrideConfig` was disabled). I'm not sure if this is the case here (but it might be). In any case though the plan is to split these into separate scripts again soon, so we might as well start here. I have a better workaround now if the same errors occur here after this change.

magnouvean (commented):
I have tested though and this looks mostly very good (more than good enough for inclusion when the changes requested are fixed)! The only two notes I have:
* It doesn't seem to work when `workspace.wallpaperPictureOfTheDay` is set (but I'm pretty sure this will be fixed when the code is moved into a separate script)
* I think when `arrangement` is an option, that it also makes sense to include an option for `alignment` (i.e. align left/align right).

modules/panels.nix
Morwdan: Definitely, thank you!

modules/workspace.nix
```diff
@@ -4,6 +4,7 @@
 let
   cfg = config.programs.plasma;
   inherit (import ../lib/wallpapers.nix { inherit lib; }) wallpaperPictureOfTheDayType wallpaperSlideShowType;
+  inherit (widgets.lib) stringIfNotNull;
```
magnouvean: I believe you need `inherit (import ./widgets/lib.nix {inherit lib; }) stringIfNotNull` here.

modules/workspace.nix
```diff
@@ -279,19 +348,40 @@ in
...
+          (if cfg.workspace.cursor.size != null then " --size ${builtins.toString cfg.workspace.cursor.size}" else "")
+          else ""}
+          ${if cfg.workspace.colorScheme != null then "plasma-apply-colorscheme ${cfg.workspace.colorScheme}" else ""}
+          ${if cfg.workspace.iconTheme != null then "${pkgs.kdePackages.plasma-workspace}/libexec/plasma-changeicons ${cfg.workspace.iconTheme}" else ""}
+        '';
+        priority = 1;
+      });
+
+      # Creates a startup script to set the desktop folder settings defined
+      # in the options under "cfg.workspace.desktop.icons".
```
magnouvean: I also don't think these comments are really needed since it's self-explanatory by the name of the script.

modules/panels.nix
Morwdan: Thanks, I tried to change this in f28bd06, but haven’t had time to check if it works. I’d also like to test if the separate script problem arises here, but I don’t know how. If I understand correctly the wallpaper got lost on the next login after changing the panels, as the wallpaper script isn’t re-run due to being unchanged while the panel script is, which also deleted the `plasma-org.kde.plasma.desktop-appletsrc` and the wallpaper settings with it. In that case I just need to change some panel settings, log off and in again and check if the desktop folder settings are still there? Or am I wrong here?

Morwdan:
(You’re reviewing faster than I can write so this was actually meant as a reply to your comment from 14 hours ago 😅)

Great, I added `alignment` and  `lockInPlace`. I also wanted to add the sorting rules, so that all settings affecting the icon placement are implemented, but I don’t know how to best handle a subset with `builtins.attrValues`, as I wanted to add a subset `sorting` like below instead of options like `sortMode`, `sortDescending` and `sortFoldersFirst`:

```nix
desktop.icons = {
  # All the other desktop icon options.
  sorting = {
    mode = lib.mkOption { … };
    descending = lib.mkOption { … };
    foldersFirst = lib.mkOption { … };
  };
};
```
Do I have to do something like this or is there a better way?
```nix
anyDesktopFolderSettingsSet = (lib.any (v: v != null) (builtins.attrValues cfg.workspace.desktop.icons)) || (lib.any (v: v != null) (builtins.attrValues cfg.workspace.desktop.icons.sorting));
# or
anyDesktopFolderSettingsSet = lib.any (v: v != null) ((builtins.attrValues cfg.workspace.desktop.icons) ++ (builtins.attrValues cfg.workspace.desktop.icons.sorting));
```
I also don’t know yet how to implement the sorting mode enum, but at least I have an idea for that. 🙈

All of today’s changes aren’t tested yet by the way, I want to do that tomorrow:

- [x] Test old options after changes in c67c5d5 and f28bd06
- [x] Test `alignment`
- [x] Test `lockInPlace`
- [ ] Test with `workspace.wallpaperPictureOfTheDay` set
- [x] Test panel update

modules/workspace.nix
Morwdan: Okay, then I’ll drop it (I tend to comment a lot in general but especially in nix as it’s still hard for me to read and understand).

modules/workspace.nix
Morwdan: Ah, because `panels.nix` has the following import in contrast to `workspace.nix`?
https://github.com/nix-community/plasma-manager/blob/8def0842f289af5d323d14219f1edccd388b195a/modules/panels.nix#L37

magnouvean:
> (You’re reviewing faster than I can write so this was actually meant as a reply to your comment from 14 hours ago 😅)
> 
> Great, I added `alignment` and `lockInPlace`. I also wanted to add the sorting rules, so that all settings affecting the icon placement are implemented, but I don’t know how to best handle a subset with `builtins.attrValues`, as I wanted to add a subset `sorting` like below instead of options like `sortMode`, `sortDescending` and `sortFoldersFirst`:
> 
> ```nix
> desktop.icons = {
>   # All the other desktop icon options.
>   sorting = {
>     mode = lib.mkOption { … };
>     descending = lib.mkOption { … };
>     foldersFirst = lib.mkOption { … };
>   };
> };
> ```
> 
> Do I have to do something like this or is there a better way?
> 
> ```nix
> anyDesktopFolderSettingsSet = (lib.any (v: v != null) (builtins.attrValues cfg.workspace.desktop.icons)) || (lib.any (v: v != null) (builtins.attrValues cfg.workspace.desktop.icons.sorting));
> # or
> anyDesktopFolderSettingsSet = lib.any (v: v != null) ((builtins.attrValues cfg.workspace.desktop.icons) ++ (builtins.attrValues cfg.workspace.desktop.icons.sorting));
> ```
> 
> I also don’t know yet how to implement the sorting mode enum, but at least I have an idea for that. 🙈
> 
> All of today’s changes aren’t tested yet by the way, I want to do that tomorrow:
> 
> * [x]  Test old options after changes in [c67c5d5](https://github.com/nix-community/plasma-manager/commit/c67c5d5e453768dea643bdf6d0c5016fd113aa5b) and [f28bd06](https://github.com/nix-community/plasma-manager/commit/f28bd0694af1abe8ee26d50c022ff408b6fb104c)
> * [x]  Test `alignment`
> * [x]  Test `lockInPlace`
> * [x]  Test with `workspace.wallpaperPictureOfTheDay` set
> * [ ]  Test panel update

I think one way of doing it would be:
```nix
anyDesktopFolderSettingsSet = 
    let
      recurse = l: lib.any (v: if builtins.isAttrs v then recurse v else v != null) (builtins.attrValues l);
    in
      recurse cfg.workspace.desktop.icons;
```
The suggestion in the quotes comment probably wouldn't work since even when none of the sorting options are set, `icons.sorting` would have value `{mode = null; descending = null; foldersFirst = null }`, which is not the same as `null` so the test wouldn't work.

magnouvean:
I've done some minor fixes now and also some testing and it seems like all my testing done so far works great. All the settings apply fine and it works with the rest of my configuration just fine. I haven't tested with panel update when `overrideConfig` is disabled, but I wouldn't worry about that, as that isn't really the responsibility of this PR (I will do some refactoring after this is merged to make sure this isn't a problem for this or other scripts). I'll let you implement sorting as you wish, and after that (and some quick testing) I reckon this is good for merging :)

modules/workspace.nix
```diff
@@ -222,11 +234,45 @@ in
...
-            defined by "arrangement", "alignment" and sorting rules (which
-            don’t have options yet) so they can’t be manually moved.
+            defined by `arrangement`, `alignment` and the `sorting` options
+            so they can’t be manually moved.
           '';
         };
 
+        sorting = {
+          mode = lib.mkOption {
+            type = with lib.types; nullOr (enum [ "manual" "name" "size" "type" "date" ]);
```
magnouvean: Use `nullOr (enum (builtins.attrNames desktopIconSortingModeId))` instead here.

modules/workspace.nix
```diff
@@ -23,14 +23,26 @@ let
     };
   };
 
+  desktopIconSortingModeId = {
+    manual = -1;
+    name   = 0;
+    size   = 1;
+    date   = 2;
+    type   = 6;
+  }.${cfg.workspace.desktop.icons.sorting.mode};
```
magnouvean: And drop `.${cfg.workspace.desktop.icons.sorting.mode}` at the end. You can convert the value in an apply statement for the option, i.e.:
```nix
sorting = {
          mode = lib.mkOption {
            type = with lib.types; nullOr (enum [ "manual" "name" "size" "type" "date" ]);
            default = null;
            example = "type";
            description = ''
              Specifies the sort mode for the desktop icons. By default they are
              sorted by name.
            '';
           apply = sortMode: if (sortMode == null) then null else desktopIconSortingModeId.${sortMode};
          };
```

magnouvean (commented):
Also would be nice if you could format with `nixpkgs-fmt`. Apart from that this works, so when this is fixed I will merge :)

modules/workspace.nix
Morwdan: Cool, I haven’t encountered `apply` before. I couldn’t find a lot of documentation I could understand so I don’t really know how it works, but [these][1] [two][2] links make it look like the result of the `apply` function will be the actual value of the option. If that is the case, should I also change [this line][3] like that?

```diff
-${stringIfNotNull cfg.workspace.desktop.icons.sorting.mode ''desktop.writeConfig("sortMode", ${builtins.toString desktopIconSortingModeId});''}
+${stringIfNotNull cfg.workspace.desktop.icons.sorting.mode ''desktop.writeConfig("sortMode", ${builtins.toString cfg.workspace.desktop.icons.sorting.mode});''}
```
Edit: I have changed it for now.

[1]: https://ryantm.github.io/nixpkgs/functions/library/options/#function-library-lib.options.mkOption
[2]: https://github.com/NixOS/nixpkgs/issues/16884#issuecomment-238814281
[3]: https://github.com/nix-community/plasma-manager/blob/0a076cdcaf6b718b18b41f929cfa49d658f9f2f7/modules/workspace.nix#L424

HeitorAugustoLN: Exactly like this!

Morwdan:
Sorry, I tested the changes after yesterday’s commit but ran out of time and therefore didn’t leave a comment. It seems to work for me. I have encountered these two problems with `overrideConfig` disabled…

- Non-default option values are applied correctly. If a previously applied option is explicitly set back to its default or to `null`, the previous setting persists as the script isn’t adapted to remove the setting and the config file is not deleted as it was before (if the panels script isn’t also run, that is)
- If the panels script _is_ run but the desktop script is not, all the options defined in the latter are lost

…but I guess that is what you meant in your previous comments, right?

> I haven't tested with panel update when `overrideConfig` is disabled, but I wouldn't worry about that, as that isn't really the responsibility of this PR (I will do some refactoring after this is merged to make sure this isn't a problem for this or other scripts).

> The reason why the wallpaper and panels had to be put in the same file was that when they were in separate scripts and the panel was updated, the wallpaper script also had to be run for the wallpaper to be set (when `overrideConfig` was disabled). I'm not sure if this is the case here (but it might be).

Morwdan:
> Also would be nice if you could format with `nixpkgs-fmt`.

Done :)

magnouvean:
This looks pretty good now. I'll just double check the functionality when I get home today and I'll merge after that. Thanks for the pr :)

Morwdan:
Thanks for the kind help, patience and your work on this project in general! It motivated me to move to NixOS on the desktop. 😊

Btw, would you also be interested in a pull request for the [mouse actions](https://github.com/nix-community/plasma-manager/compare/trunk...Morwdan:plasma-manager:mouseactions) (even though I don’t know if my approach is the right one)?

magnouvean:
No problem! Feel free to send in the mouse action as a pr if you think it works, it certainly fits the project 😃

PR #295, opened 2024-08-03
Make hex-conversion in input-module in nix instead of python

magnouvean:
As per suggestion by @apfelkuchen6 in #123


apfelkuchen6:
Wow, thanks for the incredibly quick fix. I had no idea that nixpkgs-lib contains `fromHexString`. That's far more elegant than what I imagined.

PR #298, opened 2024-08-05
Add new kscreenlocker appearance options

HeitorAugustoLN:
Also moved wallpaper options to `appearance`, because I will make another pull request later with options for kscreenlocker daemon


magnouvean:
Great, works exactly as it should :)

PR #309, opened 2024-08-10
Add mouse actions

Morwdan:
This PR adds  `Mouse Actions` from the `Desktop Folder Settings`. It works in my setup but it has a number of limitations:

- Only for "buttons" are currently supported: the left, middle and right mouse buttons as well as vertically scrolling. There is also horizontal scroll which I can only input with modifiers by holding Alt (and optionally additional modifiers) and scrolling vertically 🤔. I wasn’t sure if I should add it. All additional buttons on my G502 are not recognized.
- Modifiers are not supported. Mainly because I don’t need them but also because that would likely necessitate a different approach\* which I can’t implement. Some modifier-button-combinations don’t work (e. g. `Ctrl+Vertical-Scroll` and `Shift+Vertical-Scroll`; but `Shift+Ctrl+Vertical-Scroll` does). Modifiers that could be used are at least Ctrl, Shift, Alt, Meta and combinations of them.
- The action options, which are available for some actions like the context menu, are not supported. I haven’t looked into them.
- Some button-action-combinations seem to be incompatible (e. g. scrolling and context menu), but they can also be entered in the settings so I don’t know if we should try to catch that.
- There may be two logins required for the settings to take effect, as the settings seem to be read from the config file only on login so the changes made by the script run after the first login aren’t picked up.

Additionally I’m not confident in the way the actions are written to the config file as detailed in the temporary comments. I [haven’t found][1] a native way to write the `ActionPlugins` sections, so I’m loading the file as `ConfigFile` and write the actions inspired by [this post][2]. I’m currently leaning towards hardcoding `[ActionPlugins][0]` if you have no objections.

\* Maybe a list containing a set per action with options for the button, modifiers and action instead of the current options? Then there probably needs to be a check that every modifier-button-combinations is unique.

[1]: https://develop.kde.org/docs/plasma/scripting/
[2]: https://discuss.kde.org/t/plasma-api-scripting-set-mouse-actions-right-button-properties-in-layout-js-of-global-theme/257

### DIFF
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -49,7 +49,7 @@ let
     cfg.workspace.iconTheme != null);
 
   # Becomes true if any option under "cfg.workspace.desktop" is set to something other than null.
-  anyDesktopFolderSettingsSet =
+  anyDesktopSettingsSet =
     let
       recurse = l: lib.any (v: if builtins.isAttrs v then recurse v else v != null) (builtins.attrValues l);
     in
@@ -455,7 +455,7 @@ in
         priority = 1;
       });
 
-      desktopScript."set_desktop_folder_settings" = (lib.mkIf anyDesktopFolderSettingsSet {
+      desktopScript."set_desktop_folder_settings" = (lib.mkIf anyDesktopSettingsSet {
         text = ''
           // Desktop folder settings
           let allDesktops = desktops();
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -48,12 +48,15 @@ let
     cfg.workspace.lookAndFeel != null ||
     cfg.workspace.iconTheme != null);
 
-  # Becomes true if any option under "cfg.workspace.desktop" is set to something other than null.
+  # Becomes true if any option under "cfg.workspace.desktop.icons" is set to something other than null.
   anyDesktopFolderSettingsSet =
     let
       recurse = l: lib.any (v: if builtins.isAttrs v then recurse v else v != null) (builtins.attrValues l);
     in
-    recurse cfg.workspace.desktop;
+    recurse cfg.workspace.desktop.icons;
+
+  # Becomes true if any option under "cfg.workspace.desktop.mouseActions" is set to something other than null.
+  anyDesktopMouseActionsSet = lib.any (v: v != null) (builtins.attrValues cfg.workspace.desktop.mouseActions);
 
   splashScreenEngineDetect = theme: (if (theme == "None") then "none" else "KSplashQML");
 in
@@ -470,8 +473,12 @@ in
             ${stringIfNotNull cfg.workspace.desktop.icons.sorting.mode ''desktop.writeConfig("sortMode", ${builtins.toString cfg.workspace.desktop.icons.sorting.mode});''}
             ${lib.optionalString (cfg.workspace.desktop.icons.sorting.descending == true) ''desktop.writeConfig("sortDesc", true);''}
             ${lib.optionalString (cfg.workspace.desktop.icons.sorting.foldersFirst == false) ''desktop.writeConfig("sortDirsFirst", false);''}
-          }
+          }'';
+        priority = 3;
+      });
 
+      desktopScript."set_desktop_mouse_actions" = (lib.mkIf anyDesktopMouseActionsSet {
+        text = ''
           // Mouse actions
           // This is pretty hacky, there has to be a better way to do this.
           // I don’t even know if that works in every setup or just in my specific case,
@@ -504,7 +511,7 @@ in
           ${stringIfNotNull cfg.workspace.desktop.mouseActions.rightClick ''actionPluginSubSection.writeEntry("RightButton;NoModifier", "${cfg.workspace.desktop.mouseActions.rightClick}");''}
           ${stringIfNotNull cfg.workspace.desktop.mouseActions.verticalScroll ''actionPluginSubSection.writeEntry("wheel:Vertical;NoModifier", "${cfg.workspace.desktop.mouseActions.verticalScroll}");''}
         '';
-        priority = 3;
+        priority = 4;
       });
     };
 
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -474,6 +474,10 @@ in
           // differently on two monitors, the UI shows which actions were defined on
           // which monitor, their corresponding settings are all listed under the same
           // section though, in my case [ActionPlugins][0].
+          // Edit: After logout and login the settings in the config file seem to
+          // apply to both monitors and their respective menus represent this config.
+          // So the changes made in one monitor’s desktop settings are kept in memory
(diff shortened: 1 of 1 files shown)


modules/workspace.nix
```diff
@@ -31,18 +31,29 @@ let
...
     cfg.workspace.iconTheme != null);
 
-  # Becomes true if any option under "cfg.workspace.desktop.icons" is set to something other than null.
+  # Becomes true if any option under "cfg.workspace.desktop" is set to something other than null.
   anyDesktopFolderSettingsSet =
     let
       recurse = l: lib.any (v: if builtins.isAttrs v then recurse v else v != null) (builtins.attrValues l);
     in
-    recurse cfg.workspace.desktop.icons;
+    recurse cfg.workspace.desktop;
```
magnouvean: Since it now includes mouseactions I think this should be renamed. `anyDesktopSettingsSet` could maybe work (or something more fitting if found).

modules/workspace.nix
```diff
@@ -426,6 +471,38 @@ in
...
+          // recreated on login with it’s defaults, which then are active. It looks
+          // like one could just hard code [0], but what is the purpose of [1] then?
+          let actionPluginSubSectionId = Array.from(actionPluginSubSections).sort()[0]
+          let actionPluginSubSection = ConfigFile(configFile, actionPluginSubSectionId)
+          ${stringIfNotNull cfg.workspace.desktop.mouseActions.leftClick ''actionPluginSubSection.writeEntry("LeftButton;NoModifier", "${cfg.workspace.desktop.mouseActions.leftClick}");''}
+          ${stringIfNotNull cfg.workspace.desktop.mouseActions.middleClick ''actionPluginSubSection.writeEntry("MiddleButton;NoModifier", "${cfg.workspace.desktop.mouseActions.middleClick}");''}
+          ${stringIfNotNull cfg.workspace.desktop.mouseActions.rightClick ''actionPluginSubSection.writeEntry("RightButton;NoModifier", "${cfg.workspace.desktop.mouseActions.rightClick}");''}
+          ${stringIfNotNull cfg.workspace.desktop.mouseActions.verticalScroll ''actionPluginSubSection.writeEntry("wheel:Vertical;NoModifier", "${cfg.workspace.desktop.mouseActions.verticalScroll}");''}
```
magnouvean: It looks like this section is generated even when all the mouseActions are set to `null`. In my opinion it would be better if you check if any of the mouseActions is non-null (for example in a variable `anyDesktopMouseActionsSet`) and then keep the old `anyDesktopFolderSettingsSet`. Then you can create the two scripts only when `anyDesktopMouseActionsSet`/`anyDesktopFolderSettingsSet` is true.

modules/workspace.nix
Morwdan: The name is based on the settings window that contains the settings for the desktop icons and the mouse actions, so I think it is still _correct_ in a way, but I also think `Desktop Settings` would be a better name for this window, so I’m not at all against this change. It breaks with Plasma terminology, but other parts of this PR do to (context menu vs. standard menu).



Edit: sorry, should have read the next comment before replying 😅

modules/workspace.nix
```diff
@@ -31,18 +31,29 @@ let
...
+  mouseActions = {
+    applicationLauncher = "org.kde.applauncher";
+    contextMenu = "org.kde.contextmenu";
+    paste = "org.kde.paste";
+    switchActivity = "switchactivity";
+    switchVirtualDesktop = "org.kde.switchdesktop";
+    switchWindow = "switchwindow";
+  };
+
+  mouseActionNames = lib.types.enum (builtins.attrNames mouseActions);
```
magnouvean: I would rename to mouseActionEnum (or mouseActionNamesEnum) since this is an enum type, and the name indicates more of a list (to me anyway).

magnouvean:
I tried testing this and it did not work on my end. I have the following:
```nix
desktop.mouseActions = {
  leftClick = "applicationLauncher";
  rightClick = "switchWindow";
  middleClick = "switchVirtualDesktop";
  verticalScroll = "contextMenu";
};
```
And when I run the script, the defaults are still applied (I've also tried to log in and out to no avail). Maybe there is some other way we need to do this other than using writeEntry for it to work. Fwiw I am using `overrideConfig` so that could be what's causing the problems.

Morwdan:
> And when I run the script, the defaults are still applied (I've also tried to log in and out to no avail). Maybe there is some other way we need to do this other than using writeEntry for it to work. Fwiw I am using `overrideConfig` so that could be what's causing the problems.

I haven’t tested it with `overrideConfig` so that might be the case…

magnouvean:
I think I've found the problem on my end at the very least. It works if you just hard-code to `[ActionsPlugins][0]` but doesn't work when it writes to `[ActionsPlugins][1]`. Also it seems like I need to restart the plasma-shell afterwards, but if this is the case it's not such a big deal as we could just add `plasma-plasmashell` to the `restartServices` of the desktop-script.

Morwdan:
`[ActionsPlugins][1]` doesn’t work for me either, as mentioned in the comment. But if that is used in your case, does that mean `[ActionsPlugins][0]` is missing in your file?

Restarting `plasma-plasmashell` would then probably also get rid of the need for a second login in non-`overrideConfig`-setups, thanks :)

magnouvean:
No by default I have:
```
[ActionPlugins][0]
MiddleButton;NoModifier=org.kde.paste
RightButton;NoModifier=org.kde.contextmenu
```
, but this can just be overwritten.

Morwdan:
That’s also what I have by default, but in my case it _is_ overwritten. Anyway, I have hard-coded the section right now and will test it a bit.

magnouvean (approved):
Works good for me now

Morwdan:
For me too, except that previously defined settings are not deleted when their options are removed (with `overrideConfig` disabled), but that is probably irrelevant as it is the same problem my previous PR [has](https://github.com/nix-community/plasma-manager/pull/292#issuecomment-2274272846).

For whatever reason in one of my tests the changes in the plasma-manager options was directly written to the config file after applying the NixOS configuration, without a new login required (the changes didn’t take effect though, the login was still required for that). I couldn’t reproduce this though. But either way, one login is now sufficient.

HeitorAugustoLN:
Looks good. I will test this asap

Morwdan:
Thanks, no need to hurry though :)

HeitorAugustoLN (approved):
Works fine. I couldn't get verticalScroll actions to work (but they were set right). But it is probably because I was testing in a virtual machine

Tested using `overrideConfig`.

Morwdan:
`verticalScroll` worked for me with `switchVirtualDesktop` and `switchWindow` but had no effect in combination with other settings (probably with exception of `switchActivity`, I just have no activities to switch…), so I’d also assume/hope that this is due to your virtual machine (it did work in my test VM though, but that has nothing to say, of course).

Thanks again to both of you!

PR #332, opened 2024-08-20
Move desktop options to `desktop` module and add desktop widgets option

HeitorAugustoLN:
Help would be really appreciated.

Desktop widgets use basically the same API as panel widgets, so widgets lib can be reused, but they need to have position and size set in `ItemGeometries` and `ItemGeometriesHorizontal` with their respective IDs, so we will have to adapt it.


HeitorAugustoLN:
> set ItemGeometries and ItemGeometriesHorizontal

Using a fork of panel widgets lib for desktop widgets would be the best solution I think

HeitorAugustoLN:
Example script for desktop widget:

```js
let allDesktops = desktops();

for (let i = 0; i < allDesktops.length; i++) {
  let desktop = allDesktops[i];

  // Add the widget to the desktop
  // x = horizontal pos
  // y = vertical pos
  // w = width
  // h = height
  let widget = desktop.addWidget("org.kde.plasma.digitalclock", x, y, w, h);

}
```

HeitorAugustoLN:
I will continue working on this tomorrow, and I think I can finish it by then.

HeitorAugustoLN:
This is good to go now, `hasWidget` function needs an update, and widget specific options can be refactored to work on desktop without interfering with panel widgets using the `desktopConvert`

HeitorAugustoLN:
Tested with:
```nix
programs.plasma.workspace.desktop.widgets = [
  {
    name = "org.kde.plasma.digitalclock";
    position = { horizontal = 51; vertical = 100; };
    size = { width = 250; height = 250; };
    config.Appearance.showDate = false;
  }
  {
    name = "plasmusic-toolbar";
    position = { horizontal = 51; vertical = 300; };
    size = { width = 250; height = 400; };
    config.General.desktopWidgetBg = 4;
  }
];
```

magnouvean:
This does look quite promising. I will try to test this in the weekend. One quick thing though is that I think we should move this into a separate file from `panels.nix` (`desktop.nix` perhaps). That would mean that the code would move to its own separate startup-script, but I don't think this will be an issue (bugfixes related to this can be fixed separately, as has been planned for some weeks now).

HeitorAugustoLN:
Ok. Now I got widget-specific options to work. Will add position and size options for other widgets

HeitorAugustoLN:
> One quick thing though is that I think we should move this into a separate file from `panels.nix` (`desktop.nix` perhaps).

@magnouvean I think I can do this in this PR, also rename from `programs.plasma.workspace.desktop` to `programs.plasma.desktop`?

magnouvean:
Yeah I think that makes sense to do

HeitorAugustoLN:
I think this PR is good to merge, I will just wait for you to test. From my testing it is working perfectly fine.

magnouvean:
Looks good. I'll test on saturday probably :)

HeitorAugustoLN:
I've noticed that desktop widgets can be misplaced sometimes on startup. But that is a plasma issue, since it happened sometimes when I was testing the script with `plasma-interactiveconsole` before using it here.

HeitorAugustoLN:
It actually places in the right place, but jump to a random place after it.

magnouvean (approved):
I've tested a bit and this looks good, very impressive! I'll let you merge if you for some reason want to make some more adjustments, but all looks good to me :)

HeitorAugustoLN:
> I've tested a bit and this looks good, very impressive! I'll let you merge if you for some reason want to make some more adjustments, but all looks good to me :)

Thanks for testing!

PR #333, opened 2024-08-20
Fix data type for icon-tasks

SpringerJack:
Believe it or not: of the four filter options for the icon-tasks widget the `showOnlyMinimized` option is in fact not a bool, but an integer according to [this](https://invent.kde.org/plasma/plasma-desktop/-/blob/02726a1ca0189f71f46513078cbc82af9daa9425/applets/taskmanager/package/contents/config/main.xml). Note that the default value is set to `false` regardless which seems to be interpreted as `0` (just as true is). 

While this feels like a bug and will be reported the module in `icon-tasks.nix` for the time being does not have the desired effect when setting `behavior.showTasks.onlyMinimized = true`.


HeitorAugustoLN:
Nice spot! Could you refactor the option instead of applying the right value in the convert?

HeitorAugustoLN:
Also filled a bug report in https://bugs.kde.org/show_bug.cgi?id=491950

PR #336, opened 2024-08-22
Fix typing issues in write_config.py

luochen1990:
Add & fix type annotations so that it pass the pylance / pyright strict mode type checking.

`Any` is used for some case, might be removed and replaced by `TypedDict` if there are more information.


magnouvean:
Looks good. I'll merge right away!

PR #355, opened 2024-09-07
Lots of refactoring around startup-scripts

magnouvean:
Added a `runAlways` option for startup-scripts and desktop-scripts which is intended to be used for scripts which must run at every login. Also split the wallpapers and panels script into different scripts and not the same as was the case before (even though it wasn't really intuitive).
Furthermore, since it's a known problem that running only some of the desktop-scripts may mess up the actions of other desktop-scripts (this was the reason why panels and wallpapers were pushed into a single script in the first place), this PR makes it so that if one of the desktop-scripts need to run (for example due to one of the desktop-scripts having changed contents), all the desktop-scripts will now run together.
There's also some increased logging in the autostart script now.

I haven't done extensive testing so some more of that will be needed before merging. When I'll be able to do so I'm not sure, but hopefully not too long. If anyone else has the time to test that would of course be appreciated as always. With that said, from my early tests it seems to look quite good.

### DIFF
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -154,7 +154,7 @@ let
     };
   });
 
-  anyPanelSet = ((builtins.length cfg.panels) > 0);
+  anyPanelSet = (builtins.length cfg.panels) > 0;
 in
 {
   imports = [
--- a/modules/panels.nix
+++ b/modules/panels.nix
@@ -5,10 +5,7 @@
 } @ args:
 let
   cfg = config.programs.plasma;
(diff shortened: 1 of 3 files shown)


HeitorAugustoLN (approved):
From my testing it is working flawlessy.

HeitorAugustoLN:
Nice work on the refactor! It is looking better than before.

magnouvean:
Great! I'll just do a little more testing in one of the coming days and we hopefully can merge somewhat soon.

I think the codebase looks a little better, and is more extensible after this refactoring :)

magnouvean:
From my testing it now works fine, so I think it should be good to merge.

PR #362, opened 2024-09-12
Add KXMLGUI support and add konsole options for toolbar customization

### LINKED ISSUE #327: adding kxmlgui5 support?
I probed `~/.config`, and it turns out Konsole's `Settings -> Configure Keyboard Shortcuts` stores its data in a completely different file & format than most Plasma stuff. It dumps it under `~/.local/share/kxmlgui5/konsole/konsoleui.rc`.

This is especially weird, since I use Plasma 6 -- but it seems like a _lot_ of KDE Plasma applications store their keybindings this way. Documentation can be found [here](https://api.kde.org/frameworks/kxmlgui/).

An example from my machine's current (test) config can be found [here](https://gist.github.com/crabdancing/8ba5da4577ab5bf25660c314d53b77ce).

HeitorAugustoLN:
Closes #327

This pull request adds support for KXMLGUI configurations, adding support for Konsole shortcuts and toolbar. Additionally, the modified files have been formatted using `nixfmt-rfc-style`, aligning with the upcoming adoption of this formatter as the default for Nix. Since `nixpkgs-fmt` has been archived in favor of `nixfmt-rfc-style`, I think we should gradually apply this formatting to other files across the project. Thoughts?

### DIFF
--- a/lib/kxmlgui.nix
+++ b/lib/kxmlgui.nix
@@ -1,0 +1,226 @@
+{ lib, ... }:
+let
+  boolToString = bool: if bool then "1" else "0";
+
+  propertyType = lib.types.submodule {
+    options = {
+      name = lib.mkOption {
+        type = lib.types.str;
+        description = "The name of the property.";
+      };
+      shortcut = lib.mkOption {
+        type = lib.types.str;
+        description = "The shortcut of the property.";
+      };
+    };
+  };
+
+  actionPropertiesType = lib.types.submodule {
+    options = {
+      scheme = lib.mkOption {
+        type = lib.types.str;
+        description = "The scheme of the action properties.";
+      };
+      properties = lib.mkOption {
+        type = lib.types.listOf propertyType;
+        description = "The properties of the action properties.";
+      };
+    };
+  };
+
+  toolbarType = lib.types.submodule {
+    options = {
+      alreadyVisited = lib.mkOption {
+        type = with lib.types; nullOr bool;
+        default = null;
+        example = true;
+        description = "Whether the toolbar has already been visited.";
+        apply = value: if value != null then boolToString value else null;
+      };
+      noMerge = lib.mkOption {
+        type = with lib.types; nullOr bool;
+        default = null;
+        example = true;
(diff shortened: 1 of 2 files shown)


modules/apps/konsole.nix
```diff
@@ -121,27 +136,41 @@ in
...
+        description = ''
+          The toolbar of Konsole.
+        '';
+      };
+      session = lib.mkOption {
+        type = lib.types.nullOr kxmlguiType;
+        default = null;
+        description = ''
+          The toolbar of Konsole sessions.
+        '';
```
magnouvean, line 159: Some example properties could maybe be good here

lib/kxmlgui.nix
```diff
@@ -0,0 +1,226 @@
...
+        type = with lib.types; nullOr bool;
+        default = null;
+        example = true;
+        description = "Whether the menu bar should not be merged.";
+        apply = value: if value != null then boolToString value else null;
+      };
+      menus = lib.mkOption {
+        type = lib.types.listOf menuType;
+        description = "The menus of the menu bar.";
+      };
```
magnouvean, line 110: Could this not just be a submodule of `listOf itemType`, so instead of having something like
```
menus = [
  {
    name = "name";
    items = ["item1" ...];
  }
]
```
you could have
```
menus."name" = ["item1" ...];
```

magnouvean (commented):
The code looks good and I've done some very limited testing and the output files look as expected.

I think reformatting with `nixfmt-rfc-style` makes a lot of sense. We could do this gradually every time we modify the code, or we could do it all in one pr. I think both options are fine, but I'd maybe prefer to do it all in one pr.

Also we should perhaps add `kxmlgui5` to `defaultResetFiles` so they are reset when using `overrideConfig`. This will require some minor changes though as it's located in a separate directory, but still shouldn't be a lot of work.

HeitorAugustoLN:
I think the current approach is not that good, I think I will implement something in a script for it instead.

magnouvean:
I see that. The code would maybe be better and more extensible as well. The implementation could be very similar to the write-config script.

PR #364, opened 2024-09-15
Fix Window Rules Matching by Window Role

Green-D-683:
The Attribute window-role in kwinrulesrc is stored as windowrole (found using rc2nix). This adds an extra override to matchNameMap to account for this. Until this is merged, kwin rules cannot use Window Role Matching.


HeitorAugustoLN (approved):
Works flawlessly! Thanks for your contribution.

PR #366, opened 2024-09-21
change powerdevil.*.idleTimeOut to idleTimeout to maintain consistency

meirisoda:
I noticed that the option for 
```nix
programs.plasma.powerdevil = {
    AC.idleTimeOut 
    battery.idleTimeOut
};
```
uses `TimeOut` instead of `Timeout` like every other timeout option in powerdevil. This is not consistent with the rest of powerdevil and may cause confusion (as it did with me).

### DIFF
--- a/modules/powerdevil.nix
+++ b/modules/powerdevil.nix
@@ -129,7 +129,7 @@ let
         example = false;
         description = "Enable or disable screen dimming.";
       };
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN (approved):
Nice spot! Thanks for your contribution :)

PR #368, opened 2024-09-22
Ensure power actions are correctly applied in powerdevil module

### LINKED ISSUE #367: bug: powerdevil: some power actions will not apply to the configuration file
https://github.com/nix-community/plasma-manager/blob/7fb80fea2373c3cc9f05a84204ad0b3233464b17/modules/powerdevil.nix#L11
https://github.com/nix-community/plasma-manager/blob/7fb80fea2373c3cc9f05a84204ad0b3233464b17/modules/powerdevil.nix#L18
https://github.com/nix-community/plasma-manager/blob/7fb80fea2373c3cc9f05a84204ad0b3233464b17/modules/powerdevil.nix#L23
https://github.com/nix-community/plasma-manager/blob/7fb80fea2373c3cc9f05a84204ad0b3233464b17/modules/powerdevil.nix#L30
Setting values to null will not modify the configuration file at all, only the current settings will be retained.

HeitorAugustoLN:
Closes #367

Found these values looking into powerdevil code

From my testing should not cause any issues

### DIFF
--- a/modules/powerdevil.nix
+++ b/modules/powerdevil.nix
@@ -2,32 +2,34 @@
 let
   cfg = config.programs.plasma;
 
+  # Values can be found at:
+  # https://github.com/KDE/powerdevil/blob/master/daemon/powerdevilenums.h
   powerButtonActions = {
     nothing = 0;
     sleep = 1;
     hibernate = 2;
     shutDown = 8;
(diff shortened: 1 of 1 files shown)


magnouvean (approved):
Looks good!

PR #371, opened 2024-09-24
Fix true false regex

usama8800:


HeitorAugustoLN:
Thanks for your contribution again :)

PR #379, opened 2024-10-10
Various Additions to PowerDevil Configuration

Green-D-683:
This PR includes various additions to the plasma-manager powerdevil module - for configuring the KDE Plasma Power Management Settings:

- Display Brightness per power mode
- Power Profile per Power Mode
- Battery Level Management
  - Low/Critical Power Levels
  - Critical Power Action

There are also a couple of bugfixes:

- Display Dimming Idle Timeout was not being set properly - it should be -1 if dimming is disabled
- Lid Closing actions used "shutdown" in place of the "shutDown" used by all other actions, for the sake of consistency, this option was added alongside "shutdown" - this should probably be a replacement, but I wasn't sure how to write the mkRenamedOption or Assertion to give a reasonable warning as to the replacement having occurred.
-

### DIFF
--- a/modules/input.nix
+++ b/modules/input.nix
@@ -478,7 +478,9 @@ in
         Use.value = true;
         LayoutList.value = strings.concatStringsSep "," (map (l: l.layout) cfg.input.keyboard.layouts);
         VariantList.value = strings.concatStringsSep "," (map (l: l.variant) cfg.input.keyboard.layouts);
-        DisplayNames.value = strings.concatStringsSep "," (map (l: l.displayName) cfg.input.keyboard.layouts);
+        DisplayNames.value = strings.concatStringsSep "," (
+          map (l: l.displayName) cfg.input.keyboard.layouts
+        );
       };
     })
     (mkIf (cfg.input.keyboard.options != null) {
--- a/modules/powerdevil.nix
+++ b/modules/powerdevil.nix
@@ -178,13 +178,21 @@ let
         nullOr (enum [
           "performance"
           "balanced"
-          "power-saver"
+          "powerSaving"
         ]);
       default = null;
-      example = "power-saver";
+      example = "powerSaving";
       description = ''
         The Power Profile to Enter in this mode
       '';
+      apply =
+        profile:
+        if profile == null then
+          null
+        else if profile == "powerSaving" then
+          "power-saver"
+        else
+          profile;
     };
   };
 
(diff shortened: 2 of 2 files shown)


HeitorAugustoLN:
Didn't test it yet, but I just noticed there some 4 spaces indent in the file. Could you please format it with `nix develop && nix fmt`?

Green-D-683:
Not sure why that closed, but that should be formatted properly now

HeitorAugustoLN (approved):
Looks good to me! Did some minor changes to ensure consistency. But I'll let @magnouvean merge this. Since I don't really have a laptop, so I can't test some of these options

magnouvean:
I'll try to test it this weekend. Sorry for the longer delays here recently, I've been really busy with work lately.

magnouvean:
Mostly looks very good from my testing, but with the displayBrightness I see that when this option is set the "change screen brightness" toggle is still disabled by default. In my opinion this should automatically be enabled when this value is set, since the changing of the display brightness won't have any effect until this toggle is turned on. This should be as simple as setting:
```
UseProfileSpecificDisplayBrightness=true
```
in the powerdevilrc. Alternatively it could be added as a separate option if you think it makes sense to configure these two options independently of each other (I'd be fine with either solution tbh). Once this is fixed I think everything is good.

tucho:
I just opened a PR with changes to powerdevil, changes I had been working on for several days. I hadn't seen that this PR was open. I swear it was not an attempt to compete, I opened it from the fork on my profile, without looking at the project's PR list.

magnouvean:
That's fine, your pr seems more general while this is a little more specific, so I think we can merge this first when things are fixed and then resolve conflicts and test/merge yours @tucho :)

tucho:
Nice! Thank you @magnouvean for your reply.

Green-D-683 (commented):
> Mostly looks very good from my testing, but with the displayBrightness I see that when this option is set the "change screen brightness" toggle is still disabled by default. In my opinion this should automatically be enabled when this value is set, since the changing of the display brightness won't have any effect until this toggle is turned on. This should be as simple as setting:
```
UseProfileSpecificDisplayBrightness=true
```
in the powerdevilrc. Alternatively it could be added as a separate option if you think it makes sense to configure these two options independently of each other (I'd be fine with either solution tbh). Once this is fixed I think everything is good. 

Thank you, I must have missed that when I was developing this, as mine would have been set previously. My config is also partially ported from Plasma 5, so some of the values are still in the old file location, and I will have missed this when I checked that everything in my PowerDevilrc was available as an option.

magnouvean:
It works fine now. I'll merge right away, thanks for the contribution :)

PR #384, opened 2024-10-14
Powerdevil revamp

tucho:
I tried to cover all the options present in the configuration screen. I also made some changes to the code structure with the intention of making it more uniform and organized.

As for the option names, my idea was to mirror the texts of the options on the configuration screen. I think that even with longer names, it will be easier for the end user to assemble their configuration from the screens.

Please consider my contribution as a suggestion and not as an attempt to change for the sake of change. We can and should discuss these changes and, if necessary, reach a compromise in terms of nomenclature and organization.

One more thing: don't be scared by the number of commits, it's just my way of working. I suggest that you analyze the changes based on the final code.

### DIFF
--- a/modules/powerdevil.nix
+++ b/modules/powerdevil.nix
@@ -90,21 +90,22 @@ let
           If enabled, the lid action will be inhibited when an external monitor is connected.
         '';
       };
+
+      whenSleepingEnter = lib.mkOption {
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN:
I think this mostly looks good, besides the reformatting of lists, because they are formatted like this because of `nixfmt`. But I still need to take a better look at it.

max06:
Looking forward to this - the new display dimming options in kde are annoying. 👍🏼

PR #389, opened 2024-10-19
apps: make package installation optional

SigmaSquadron:
Users should be able to configure a system-wide installation of `{kate,ghostwriter,okular}` without needing to have a duplicate home manager package. This behaviour was even documented in Kate's `package` option, but did not work as the `nullable` tag wasn't set to true. This commit corrects this behaviour.


HeitorAugustoLN:
Really good change! Thanks for the contribution.

PR #391, opened 2024-10-19
feat(elisa): init Elisa module

SigmaSquadron:
Elisa is the KDE Music Player. It has very few options compared to most KDE apps, but can still be configured relatively easily.

### DIFF
--- a/modules/apps/default.nix
+++ b/modules/apps/default.nix
@@ -2,6 +2,7 @@
 
 {
   imports = [
+    ./elisa.nix
     ./ghostwriter.nix
     ./konsole.nix
     ./kate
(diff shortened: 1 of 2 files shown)


HeitorAugustoLN:
This is looking pretty good, I will make some edits after I finish my neovim configuration and then I'll merge it :)

SigmaSquadron:
note to self: don't do *that*

HeitorAugustoLN (approved):
LGTM! Sorry for the big delay. I'll merge right away.

SigmaSquadron:
Formatted and removed the `with lib.strings;`, as it makes the scope unclear.

HeitorAugustoLN:
> Formatted and removed the `with lib.strings;`, as it makes the scope unclear.

In that case it is not a problem, since this adds the `lib.strings` scope only to the function, so it is easier to understand, where these functions come from, but I don't mind it

HeitorAugustoLN (approved):

PR #397, opened 2024-10-19
fix(kwin): update usage of `findFirstIndex` from `lib` to `lib.lists`

HeitorAugustoLN:


SigmaSquadron:
oops. sorry about that!

PR #410, opened 2024-11-13
Add explicit iconOnly option for the taskmanager

thomas-bouvier:
Hello,

I am looking for some guidance to implement an option enabling the Icons-and-Text task manager widget. At the moment, only the Icons task manager widget seems supported by this project.

I couldn't find any documentation on the matter, and this is my first Nix contribution. Looking at [this file](https://github.com/KDE/plasma-desktop/blob/3baa5b8/applets/taskmanager/package/contents/ui/ConfigAppearance.qml#L19), it seems that the icons-only mode is activated when package `org.kde.plasma.icontasks` is used in place of `org.kde.plasma.taskmanager`. However, I'd like to see some validation of the robustness of this approach.

Also, the `iconOnly` boolean may need to be inverted to keep the default behavior to icons-only mode.

I'm currently testing to see if all the options apply to `org.kde.plasma.taskmanager`.

### DIFF
--- a/modules/widgets/icon-tasks.nix
+++ b/modules/widgets/icon-tasks.nix
@@ -240,7 +240,7 @@ in
         behavior,
         launchers,
         settings,
-        iconsOnly
+        iconsOnly,
         ...
       }:
       {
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN:
I guess this should fix the PR, the `mkBoolOption` sets a null value by default, that why the error was happening, that should do the work for `taskmanager` widget working. If you think it is good to go, you can mark the PR as ready for review and I would be happy to merge :)

thomas-bouvier:
Hello @HeitorAugustoLN, thank you for the fix, all is good on my side :)

HeitorAugustoLN (approved):

PR #411, opened 2024-11-13
Added tabBox config to kwin module

JustScreaMy:
Hello, I've just started using this and noticed a few missing options for configuring how the Task Switcher behaves.

I had some uncertainty around naming the configuration options and ran into a couple of issues with handling options that use 0/1 values (to represent different modes) instead of simple true/false booleans.

Just to explain my thought process behind handling `includeShowDesktop` and `sortMode` differently, I think that it is more likely that a new Sorting order would be added than they add more ShowDesktopModes.

Looking forward to any feedback and suggestions on how to improve the clarity and usability of these configuration options.

### DIFF
--- a/modules/kwin.nix
+++ b/modules/kwin.nix
@@ -439,6 +439,67 @@ in
       };
     };
 
+    tabBox = {
+      showTabBox = lib.mkOption {
+        type = with lib.types; nullOr bool;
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN:
I will review this soon.

Issue #419, opened 2024-11-21
Feature Request: Add Support for Plasma 6 Krohnkite Tiling Script


sidmoreoss:
The [Krohnkite Tiling Script](https://github.com/anametologin/krohnkite) has recently gained traction in the Plasma 6 community and is widely regarded as one of the best tiling scripts available for KWin.

Are there any plans to add support for the Krohnkite tiling script in Plasma 6?
Has work already started on this, or is there an initiative to include it in upcoming releases?

I am happy to contribute to integrating Krohnkite support!

miro-zamiro:
I would also really like seeing Krohnkite implemented!!!

an0t8:
This is even more important now since nixpkgs removed polonium earlier this month.  I would love to see the polonium script replaced with krohnkite in plasma-manager.

PR #420, opened 2024-11-22
feat(Krohnkite): init Krohnkite kwin script

sidmoreoss:
https://github.com/nix-community/plasma-manager/issues/419

### DIFF
--- a/modules/kwin.nix
+++ b/modules/kwin.nix
@@ -56,6 +56,10 @@ let
     names:
     builtins.listToAttrs (lib.imap1 (i: v: (lib.nameValuePair "Name_${builtins.toString i}" v)) names);
 
+  virtualDesktopIdAttrs =
+    names:
+    builtins.listToAttrs (lib.imap1 (i: v: (lib.nameValuePair "Id_${builtins.toString i}" v)) names);
+
   capitalizeWord =
     word:
     let
@@ -712,6 +716,7 @@ in
           (lib.mkIf (cfg.kwin.virtualDesktops.names != null) {
             Desktops = lib.mkMerge [
               { Number = builtins.length cfg.kwin.virtualDesktops.names; }
+              (virtualDesktopIdAttrs cfg.kwin.virtualDesktops.names)
               (virtualDesktopNameAttrs cfg.kwin.virtualDesktops.names)
             ];
           })
--- a/modules/kwin.nix
+++ b/modules/kwin.nix
@@ -948,7 +948,7 @@ in
           })
 
           (lib.mkIf (cfg.kwin.scripts.krohnkite.enable != null) {
-            Plugins.krohnkiteEnable = cfg.kwin.scripts.krohnkite.enable;
+            Plugins.krohnkiteEnabled = cfg.kwin.scripts.krohnkite.enable;
             Script-krohnkite = {
               screenGapTop = cfg.kwin.scripts.krohnkite.settings.geometry.gaps.top;
               screenGapLeft = cfg.kwin.scripts.krohnkite.settings.geometry.gaps.left;
--- a/modules/kwin.nix
+++ b/modules/kwin.nix
@@ -4,7 +4,6 @@
   pkgs,
   ...
 }:
-
 let
   cfg = config.programs.plasma;
   validTitlebarButtons = {
@@ -56,10 +55,13 @@ let
     names:
     builtins.listToAttrs (lib.imap1 (i: v: (lib.nameValuePair "Name_${builtins.toString i}" v)) names);
 
-
   virtualDesktopIdAttrs =
     number:
-    builtins.listToAttrs (map (i: (lib.nameValuePair "Id_${builtins.toString i}" "Desktop_${builtins.toString i}")) (lib.range 1 number));
+    builtins.listToAttrs (
+      map (i: (lib.nameValuePair "Id_${builtins.toString i}" "Desktop_${builtins.toString i}")) (
+        lib.range 1 number
+      )
+    );
 
   capitalizeWord =
     word:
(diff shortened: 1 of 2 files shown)


modules/kwin.nix
```diff
@@ -564,6 +571,175 @@ in
...
+        enable = lib.mkOption {
+          type = with lib.types; nullOr bool;
+          default = null;
+          example = true;
+          description = "Whether to enable Krohnkite.";
+        };
+
+        settings = {
+          geometry = {
+            gaps = {
```
SigmaSquadron: I think it would make more sense if `gaps` is itself an option that takes in a submodule as a value.

modules/kwin.nix
```diff
@@ -564,6 +571,175 @@ in
...
+                type = with lib.types; nullOr bool;
+                default = null;
+                example = true;
+                description = "Whether to enable monocle layout for krohnkite.";
+              };
+              maximize = lib.mkOption {
+                type = with lib.types; nullOr bool;
+                default = null;
+                example = true;
+                description = "Fully maximize windows (no borders, no gaps) for monocle layout.";
```
SigmaSquadron: Isn't this the same as having all gaps set to `0`?

modules/kwin.nix
```diff
@@ -564,6 +571,175 @@ in
...
+              ratio = lib.mkOption {
+                type = with lib.types; nullOr (ints.between 1 100);
+                default = null;
+                example = 1.6;
+                description = "Tile width limiting ratio for krohnkite.";
+              };
+            };
+          };
+
+          layouts = {
```
SigmaSquadron: Perhaps this should also be an option that takes in a `oneOf` string value.

modules/kwin.nix
```diff
@@ -633,7 +809,9 @@ in
 
       home.packages =
         with pkgs;
-        [ ] ++ lib.optionals (cfg.kwin.scripts.polonium.enable == true) [ polonium ];
+        [ ]
+        ++ lib.optionals (cfg.kwin.scripts.polonium.enable == true) [ polonium ]
+        ++ lib.optionals (cfg.kwin.scripts.krohnkite.enable == true) [ kdePackages.krohnkite ];
```
SigmaSquadron:
```suggestion
        home.packages = lib.optionals cfg.kwin.scripts.polonium.enable [ pkgs.polonium ] ++ lib.optionals cfg.kwin.scripts.krohnkite.enable [ pkgs.kdePackages.krohnkite ];
```

modules/kwin.nix
```diff
@@ -765,6 +947,34 @@ in
...
+              enableQuarterLayout = cfg.kwin.scripts.krohnkite.settings.layouts.quarter.enable;
+              enableSpiralLayout = cfg.kwin.scripts.krohnkite.settings.layouts.spiral.enable;
+              enableSpreadLayout = cfg.kwin.scripts.krohnkite.settings.layouts.spread.enable;
+              enableStackedLayout = cfg.kwin.scripts.krohnkite.settings.layouts.stacked.enable;
+              enableStairLayout = cfg.kwin.scripts.krohnkite.settings.layouts.stair.enable;
+              enableThreeColumnLayout = cfg.kwin.scripts.krohnkite.settings.layouts.threeColumn.enable;
+              enableTileLayout = cfg.kwin.scripts.krohnkite.settings.layouts.tile.enable;
+              layoutPerActivity = cfg.kwin.scripts.krohnkite.settings.layouts.enableLayoutPerActivity;
+              layoutPerDesktop = cfg.kwin.scripts.krohnkite.settings.layouts.enableLayoutPerDesktop;
+            };
```
SigmaSquadron: this can probably be abstracted away in a `keyValuePair` generator.

modules/kwin.nix
HeitorAugustoLN: This wouldn't work, because the default values are `null`, and not a boolean

SigmaSquadron: Right! Sorry about that, I'm used to not having `null` in `enable` options.

modules/kwin.nix
sidmoreoss: Could you please elaborate a bit more on what you mean by that?
If possible, could you provide an example or additional context to help me better understand your idea?

modules/kwin.nix
sidmoreoss: could you provide an example or additional context

modules/kwin.nix
sidmoreoss: do you mean to remove the geometry key altogether so that the structure becomes settings.gaps.top?

sidmoreoss: or do you mean something like this 
image

modules/kwin.nix
sidmoreoss: No, maximize for monocle layout removes all borders (this includes window decorations) and gaps and makes the active window fullscreen

modules/kwin.nix
SigmaSquadron: There are generators in the Nixpkgs `lib` and `pkgs-lib` that can create these attrsets (or strings to be written to files directly, if desired). You can use those generators to avoid having keys for *every single option* declared here - the generator will instead dynamically assign the right keys to the right values, which will work regardless of how many keys there are. A similar thing is done on NixOS and Home Manager modules, particularly the `bat` module for Home Manager, although that's building the file itself instead of an attrset.

modules/kwin.nix
SigmaSquadron: LGTM.

modules/kwin.nix
SigmaSquadron: LGTM.

modules/kwin.nix
SigmaSquadron: Also see RFC 0042.

sidmoreoss:
@HeitorAugustoLN 
Could you please let me know if there's anything else that needs to be done here?

HeitorAugustoLN:
I will review soon, I've been kinda busy these days working on some other projects. But I will set aside some time to review this when I can, shouldn't take too long.

modules/scripts/tiling/krohnkite.nix
```diff
@@ -0,0 +1,315 @@
...
+      ];
+      ignoreActivities = mkFilterOption "Disable tiling on activities" [ "Activity1" ];
+      ignoreScreens = mkFilterOption "Disable tiling on screens" [ "Screen1" ];
+      ignoreVirtualDesktops = mkFilterOption "Disable tiling on virtual desktops" [ "Desktop_1" ];
+      floatWindowsByClass = mkFilterOption "Float windows by class" [ "dialog" ];
+      floatWindowsByTitle = mkFilterOption "Float windows by title" [ "firefox" ];
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
```
SigmaSquadron, line 270: I think the condition should be something else here? Enabling `programs.plasma.enable` shouldn't require evaluating Krohnkite's `config`.

j-pap:
With Polonium being archived in November 2025, I'd love to see this get merged as an alternative.

vendion:
It seems the polonuim being archived has finally come to hit us

```
 … while calling the 'throw' builtin
┃          at /nix/store/vp1i814m5cqfzh22bl76s1k9bc8iikkv-source/pkgs/top-level/aliases.nix:1706:14:
┃          1705|   polipo = throw "'polipo' has been removed as it is unmaintained upstream"; # Added 2025-05-18
┃          1706|   polonium = throw "'polonium' has been removed, as Plasma 5 has reached end of life."; # Added 2026-05-01
┃              |              ^
┃          1707|   polypane = throw "'polypane' has been removed due to lack of maintenance in nixpkgs"; # Added 2025-06-25
┃
┃        error: 'polonium' has been removed, as Plasma 5 has reached end of life.
```

ShadowRZ:
@sidmoreoss I'm not sure why the checks failed, maybe @HeitorAugustoLN should have a look and merge this?

Dopplerian:
> @sidmoreoss I'm not sure why the checks failed, maybe @HeitorAugustoLN should have a look and merge this?

From what I can gather, it seems the problem is some minor GitHub API issue, so it shouldn't take too long to fix, I think (although I'm not sure how).

It would be nice if @HeitorAugustoLN or anyone else could fix it, given that Polonium is no longer an alternative.

ShadowRZ:
> It would be nice if @HeitorAugustoLN or anyone else could fix it, given that Polonium is no longer an alternative.

Cc @sidmoreoss maybe close and reopen it would rerun the CI?

Also it just happened that since then [`polonium` was revived](https://github.com/NixOS/nixpkgs/pull/532117), also updated to 1.2.0

j-pap:
Here's how I was able to get Krohnkite working:

```
configFile.kwinrc = {
  Plugins.krohnkiteEnabled = true;
  Script-krohnkite = {
    layoutPerDesktop = true;
    monocleLayoutOrder = 2;
    screenGapTop = 10;
    screenGapLeft = 10;
    soleWindowNoBorders = true;
    threeColumnLayoutOrder = 0; # 0 is disabled
    tileLayoutOrder = 1;
    # etc...
  };
};
```

You can find all Khronkite's existing settings in your kwinrc file under [Script-krohnkite], then just set them as attributes as shown above.

PR #439, opened 2024-12-22
deduplicated font type

Asqiir:
I noticed that we define the same type twice. I deduplicated that type.


Asqiir:
The CI piple fails at the step `nix flake check`.
This happens because the function `lib.defaultTo` was only added recently https://github.com/NixOS/nixpkgs/commit/8ba40fdd4fae1edd4d725a53c04401d1f0d8b2a5
But it is already in `nixos-unstable`. I propose that the lockfile is updated.

Asqiir:
This depends on #442, as the function `defaultTo` isn't accessible with our current `nixpkgs` version.

HeitorAugustoLN:
LGTM, but I will already do something similar on my refactor, which will fit better, without needing to be importing, but just calling it as a regular type

PR #443, opened 2024-12-24
feat(input): touchpad.scrollSpeed

aster-void:


modules/input.nix
```diff
@@ -222,6 +231,7 @@ let
         TapAndDrag = touchpad.tapAndDrag;
         TapDragLock = touchpad.tapDragLock;
         ScrollMethod = touchpad.scrollMethod;
+        ScrollFactor = touchpad.scrollSpeed;
```
bew, line 234: Why not call it `scrollFactor` like the actual config option?

aster-void: because it's also called scrollSpeed in mouse section.
I followed it because scrollFactor is a little bit unintuitive and would make it difficult to search and autocomplete

HeitorAugustoLN (approved):
I guess we forgot to add it before! Thanks!

PR #444, opened 2024-12-24
doc(input): add some examples to keyboard layout variant

aster-void:
It should include examples of layout (such as Dvorak, Colemak, etc).
"mac" isn't necessary, but it's better than not having mac in there.


HeitorAugustoLN (approved):

HeitorAugustoLN:
Thanks for your contribution!

PR #470, opened 2025-03-01
Allow setting different wallpapers for different screens

icewind1991:
Allows configuring different wallpaper for different screens by setting `programs.plasma.workspace.wallpaper` to a list of paths instead of a single path.

### DIFF
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -161,7 +161,7 @@ in
     };
 
     wallpaper = lib.mkOption {
-      type = with lib.types; nullOr path;
+      type = with lib.types; nullOr (either path (listOf path));
       default = null;
       example = lib.literalExpression ''"''${pkgs.kdePackages.plasma-workspace-wallpapers}/share/wallpapers/Kay/contents/images/1080x1920.png"'';
       description = ''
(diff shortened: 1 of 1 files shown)


modules/workspace.nix
```diff
@@ -453,34 +453,44 @@ in
               for (const desktop of allDesktops) {
                   desktop.wallpaperPlugin = "org.kde.image";
                   desktop.currentConfigGroup = ["Wallpaper", "org.kde.image", "General"];
-                  desktop.writeConfig("Image", "file://${toString cfg.workspace.wallpaper}");
                   ${
-                    lib.optionalString (cfg.workspace.wallpaperFillMode != null)
-                      ''desktop.writeConfig("FillMode", "${
-                        toString wallpaperFillModeTypes.${cfg.workspace.wallpaperFillMode}
-                      }");''
```
magnouvean: Why was this removed?

magnouvean (commented):
Just had a question about one part of the code. Has this been quickly tested when setting the value to one wallpaper (i.e. a path instead of a list of paths) to ensure this doesn't break that? Generally looks good, so should be OK for merging if this is confirmed working.

modules/workspace.nix
icewind1991: I think this accidentally got removed during a rebase, fix now

icewind1991:
> Has this been quickly tested when setting the value to one wallpaper (i.e. a path instead of a list of paths) to ensure this doesn't break that?

I've been using this on both a multi monitor setup and a single monitor setup without issues

magnouvean:
Great! LGTM

BCVerdouw:
It did seem to break using a path on a multi-monitor setup at the very least, see issue #511 and PR #512

PR #471, opened 2025-03-06
feat(cursors): Cursor launch feedback

4lon:
While I was configuring my kde setup I noticed that the settings for cursor launch feedback seemed to be missing.
image

I had a go at adding these settings to the cursorType since they are listed under cursors in kde settings, and added configuration to klaunchrc which I believe is the only place these settings have an effect. 

I'm quite new to nix so please let me know if it needs any changes or if you need me to refactor this.

### DIFF
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -36,19 +36,19 @@ let
           default = null;
           example = "Bouncing";
           description = ''
-	    The cursor feedback icon after launching an application. Valid options are Bouncing, Blinking, Static, and None.
+            	    The cursor feedback icon after launching an application. Valid options are Bouncing, Blinking, Static, and None.
           '';
         };
         taskManagerFeedback = lib.mkOption {
           type = nullOr bool;
           default = null;
-	  example = true;
+          example = true;
           description = "The feedback wheel on an application icon after launching an application from the task manager.";
         };
         animationTime = lib.mkOption {
           type = nullOr ints.positive;
           default = null;
-	  example = 5;
+          example = 5;
           description = "The duration that the cursorFeedback and taskManagerFeedback run for.";
         };
       };
@@ -156,9 +156,9 @@ in
       example = {
         theme = "Breeze_Snow";
         size = 24;
-	cursorFeedback = "Bouncing";
-	taskManagerFeedback = true;
-	animationTime = 5;
+        cursorFeedback = "Bouncing";
+        taskManagerFeedback = true;
+        animationTime = 5;
       };
       description = ''
         Submodule for configuring the cursor appearance. The theme, size, cursor feedback, task manager feedback, and animation time are configurable.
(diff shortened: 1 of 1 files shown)


modules/workspace.nix
```diff
@@ -31,6 +31,26 @@ let
...
           description = "The size of the cursor. See the System Settings app for allowed sizes for each cursor theme.";
         };
+        cursorFeedback = lib.mkOption {
+          type = nullOr str;
+          default = null;
+          example = "Bouncing";
+          description = ''
+	    The cursor feedback icon after launching an application. Valid options are Bouncing, Blinking, Static, and None.
+          '';
+        };
```
HeitorAugustoLN, line 44: Here you can define an enum of values instead:

```suggestion
        cursorFeedback = lib.mkOption {
          type = nullOr (enum [ "Bouncing" "Blinking" "Static" "None" ]);
          default = null;
          example = "Bouncing";
          description = ''
	    The cursor feedback icon after launching an application.
          '';
        };
```

HeitorAugustoLN (dismissed):
Thanks for contributing to `plasma-manager`! Also please format the code you wrote with `nixfmt-rfc-style`. I also need to check if the values are working correctly, which I will do later, since I am not using plasma right now, so I will need to set up a vm first, but everything else looks good to me

modules/workspace.nix
4lon: Thanks for your help! I've formatted the sections I wrote and updated the cursorFeedback to use an enum instead of string.

modules/workspace.nix
```diff
@@ -354,6 +380,42 @@ in
...
+                  "Blinking" = {
+                    BusyCursorSettings.Blinking = "true";
+                    BusyCursorSettings.Bouncing = "false";
+                    FeedbackStyle.BusyCursor = "true";
+                  };
+                  "Bouncing" = {
+                    BusyCursorSettings.Blinking = "false";
+                    BusyCursorSettings.Bouncing = "true";
+                    FeedbackStyle.BusyCursor = "true";
+                  };
```
magnouvean, line 413: Are the quotes really necessary here? I would think maybe for example:
```nix
{
    None = {
        BusyCursorSettings.Blinking = false;
        BusyCursorSettings.Bouncing = false;
        FeedbackStyle.BusyCursor = false;
    };
    ...
}
```

Would work? If so that should be better. Also maybe grouping on BusyCursorSettings is cleaner, i.e.:
```nix
{
    None = {
        BusyCursorSettings = {
            Blinking = false;
            Bouncing = false;
        };
        FeedbackStyle.BusyCursor = false;
    };
    ...
}
```

magnouvean (changes requested):
Generally this looks good now imo. See my one comment though. After that is cleared I reckon we can merge :)

Sorry for the late review here, I haven't had much time for plasma-manager lately due to other work.

depeo:
This would be a great feature! The bouncing cursor looks so tacky! 😄

modules/workspace.nix
4lon: Firstly, I am really sorry for how long it took me to get back to this, I haven't been using kde recently so I had to get re-acquainted hahaha. 

You're right, this is cleaner, I was pretty new to nix when wrote this. I have taken your advice and removed the quotes and grouped the BusyCursorSettings.

magnouvean (approved):
No problem, and thanks for fixing 👍. Looks good now 😄

PR #481, opened 2025-04-09
update basic test

in-a-dil-emma:
- The basic test was difficult to read.
- The basic test wasn't checking if a key set directly through configFile would get overridden by an abstraction.


magnouvean:
Cool. I don't think we've used the tests that much currently, but I don't see anything wrong here. LGTM

PR #493, opened 2025-05-25
feat(kwin/navigation-wrapping): add option for navigation-wrapping

antoniosubasic:
Extend [programs.plasma.kwin.effects.desktopSwitching](https://nix-community.github.io/plasma-manager/options.xhtml#opt-programs.plasma.kwin.effects.desktopSwitching.animation) by adding option for wrapping around virtual desktops when switching through.

native virtual desktops option - system settings

### DIFF
--- a/modules/kwin.nix
+++ b/modules/kwin.nix
@@ -209,17 +209,24 @@ in
         default = null;
         description = "Arrange desktops in a virtual cube.";
       };
-      desktopSwitching.animation = lib.mkOption {
-        type =
-          with lib.types;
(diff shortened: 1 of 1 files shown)


magnouvean:
Nice. I only saw this after #509, but the hierarchy here is probably better. I reckon we can merge this and then just revert that PR, so that the option we use is  `programs.plasma.kwin.effects.desktopSwitching.navigationWrapping`

antoniosubasic:
Sweet. Thanks

PR #494, opened 2025-05-26
Fix: Setting wallpaper path in script

Faupi:
Simply removed `file://` from the wallpaper path, as also KDE itself just uses the direct path and using this prefix makes SVG (potentially other formats as well?) not work for one reason or another. 

Tested with my personal SVG wallpaper, but I don't see how it would cause issues with others, as this matches the native Plasma 6 handling on my system:
```
Operating System: NixOS 25.05
KDE Plasma Version: 6.3.5
KDE Frameworks Version: 6.14.0
Qt Version: 6.9.0
Kernel Version: 6.12.26 (64-bit)
Graphics Platform: Wayland```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1740494361,
-        "narHash": "sha256-Dd/GhJ9qKmUwuhgt/PAROG8J6YdU2ZjtJI9SQX5sVQI=",
+        "lastModified": 1756579987,
+        "narHash": "sha256-duCce8zGsaMsrqqOmLOsuaV1PVIw/vXWnKuLKZClsGg=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "74f0a8546e3f2458c870cf90fc4b38ac1f498b17",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Faupi** [comment]
Bleh, seems like the issue might be due to something else and my change just so happened to resolve it temporarily, or it's possible that what I run into with the wallpaper is just something stacked on top, as I got it again despite leaving out the `file://` prefix:
```
plasmashell[2595]: qml: The backend got an unknown wallpaper provider type. The wallpaper will now fall back to the default. Please check your wallpaper configuration!
```

_Still, since Plasma writes the path without the prefix, maybe it's good to do it the same anyway, but I'm fine with skipping this change as well._ I'll be on the lookout for what actually causes these issues with SVG wallpapers.

Relevant sample from `~/.config/plasma-org.kde.plasma.desktop-appletsrc` for anyone interested:
```
[Containments][1]
...
plugin=org.kde.plasma.folder
wallpaperplugin=org.kde.image

...

[Containments][1][Wallpaper][org.kde.image][General]
Image=/nix/store/vzhrijwva1jj65daddhv4522gi04msq1-source/home-manager/cfgs/shared/kde-plasma/wallpaper.svg
```

PR #520, opened 2025-08-31
fix: rename homeManagerModules to homeModules

andre4ik3:
The proper output name for Home Manager modules is `homeModules`, not `homeManagerModules`. See [here][1] and [here][2] and the discussion [here][3].

Also, I changed `import inputs.nixpkgs { ... }` to `nixpkgs.legacyPackages`, to avoid creating another instance of nixpkgs (small performance optimization).

[1]: https://github.com/nix-community/home-manager/blob/99a69bdf8a3c6bf038c4121e9c4b6e99706a187a/flake-module.nix#L28-L43
[2]: https://github.com/DeterminateSystems/flake-schemas/blob/d0e74ee9a30eda4cc153b7f1e347043680834180/flake.nix#L293-L306
[3]: https://github.com/nix-community/home-manager/pull/6392#issuecomment-2633445057

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -25,11 +25,15 @@
       nixpkgsFor = forAllSystems (system: import inputs.nixpkgs { inherit system; });
     in
     {
-      homeManagerModules.plasma-manager =
+      homeModules.plasma-manager =
         { ... }:
         {
           imports = [ ./modules ];
         };
+
+      homeManagerModules = inputs.nixpkgs.lib.warn ''
+        plasma-manager: homeManagerModules has been renamed to homeModules
+      '' self.homeModules;
 
       packages = forAllSystems (
         system:
--- a/examples/homeManagerFlake/flake.nix
+++ b/examples/homeManagerFlake/flake.nix
@@ -34,7 +34,7 @@
         pkgs = import nixpkgs { inherit system; };
 
         modules = [
-          inputs.plasma-manager.homeManagerModules.plasma-manager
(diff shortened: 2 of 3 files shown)


flake.nix
```diff
@@ -22,15 +22,19 @@
       forAllSystems = inputs.nixpkgs.lib.genAttrs supportedSystems;
 
       # Attribute set of nixpkgs for each system:
-      nixpkgsFor = forAllSystems (system: import inputs.nixpkgs { inherit system; });
```
HeitorAugustoLN, line 25: Unrelated change for this PR, can be done in a separate one. Also not sure about the performance improvement on that

andre4ik3: I thought it was too minor for a separate PR, but ok, I'll make another one, 1 sec

HeitorAugustoLN: yeah, it is, but i am not sure if I want this change for now, I will take a look on it later.

HeitorAugustoLN (approved):
Thanks for this contribution :)

andre4ik3:
(Btw the CI is failing because the Nix version is too old to evaluate Nixpkgs)

HeitorAugustoLN:
> (Btw the CI is failing because the Nix version is too old to evaluate Nixpkgs)

yeah just read it

andre4ik3:
Here's a CI fix https://github.com/nix-community/plasma-manager/pull/523

PR #523, opened 2025-08-31
fix: use determinate nix installer in CI

andre4ik3:
Fixes CI failing due to too old Nixpkgs


HeitorAugustoLN:
Use `DeterminateSystems/nix-installer-action` instead. This doesn't install regular nix

andre4ik3:
Oh right sorry I skim read

CleanShot 2025-08-31 at 13 33 17@2x

andre4ik3:
The determinate installer now plans to drop support for normal Nix, so I changed it to just update the cachix action

PR #530, opened 2025-09-19
fix: prevent wallpapers from being garbage collected

antoniosubasic:
Whilst configuring wallpapers ([kscreenlocker](https://nix-community.github.io/plasma-manager/options.xhtml#opt-programs.plasma.kscreenlocker.appearance.wallpaper) and [workspace](https://nix-community.github.io/plasma-manager/options.xhtml#opt-programs.plasma.workspace.wallpaper)) on my system, I noticed that everything appeared to work after a rebuild, but the wallpapers disappeared after garbage collecting and rebooting.

Previously, wallpapers were copied into the nix store and their paths written directly to the KDE configuration files, resulting in them lying around unreferenced in the nix store and eventually being removed by the garbage collector.

This change wraps wallpapers with `builtins.path { path = ...; }`, which imports them into the nix store as declared dependencies. This ensures that the wallpapers are part of the system closure and thus retained by the garbage collector as long as it is referenced by the system configuration.

I'm not sure whether this is the most idiomatic nix solution, but I've found it to work for me. I'm open to feedback and happy to adjust if there’s a better approach to this problem.

### DIFF
--- a/modules/kscreenlocker.nix
+++ b/modules/kscreenlocker.nix
@@ -203,6 +203,7 @@ in
         (lib.mkIf (cfg.kscreenlocker.appearance.wallpaper != null) {
           Greeter.WallpaperPlugin = "org.kde.image";
           "Greeter/Wallpaper/org.kde.image/General".Image = builtins.path {
+            name = "kscreenlocker-wallpaper";
             path = cfg.kscreenlocker.appearance.wallpaper;
           };
         })
--- a/modules/kscreenlocker.nix
+++ b/modules/kscreenlocker.nix
@@ -202,9 +202,9 @@ in
       lib.mkMerge [
         (lib.mkIf (cfg.kscreenlocker.appearance.wallpaper != null) {
           Greeter.WallpaperPlugin = "org.kde.image";
-          "Greeter/Wallpaper/org.kde.image/General".Image = (
-            builtins.toString cfg.kscreenlocker.appearance.wallpaper
-          );
+          "Greeter/Wallpaper/org.kde.image/General".Image = builtins.path {
+            path = cfg.kscreenlocker.appearance.wallpaper;
+          };
         })
         (lib.mkIf (cfg.kscreenlocker.appearance.wallpaperPictureOfTheDay != null) {
           Greeter.WallpaperPlugin = "org.kde.potd";
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -457,46 +457,55 @@ in
                     if (builtins.typeOf cfg.workspace.wallpaper) == "list" then
                       ''
                         let images = [
-                        ${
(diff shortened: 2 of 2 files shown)


sgsunder:
I had the same problem as well, but your PR helped me find a more direct solution to this.

Would get garbage collected:
```nix
programs.plasma.workspace.wallpaper = ./image.png;
```

Works across garbage collection:
```nix
programs.plasma.workspace.wallpaper = builtins.path { path = ./image.png; };
```

Can still see the PR being a good idea though, to avoid this confusion.

antoniosubasic:
I'm glad I was able to help.

I suggested this PR because the option `programs.plasma.workspace.wallpaper` expects the user to just set the wallpaper desired and the user shouldn't (in my opinion) have to care about or handle garbage collection.

HeitorAugustoLN (approved):
Thanks!

axgfn:
This change broke my config by no longer allowing me to reference a wallpaper in my home directory. I realize that makes my configuration impure, but I don't plan on putting the image in the nix store. Unsure how to work around this at the moment - for now I'm just pinning my plasma-manger input to the commit before this was merged.

NovaViper:
> This change broke my config by no longer allowing me to reference a wallpaper in my home directory. I realize that makes my configuration impure, but I don't plan on putting the image in the nix store. Unsure how to work around this at the moment - for now I'm just pinning my plasma-manger input to the commit before this was merged.

I made a separate [github repo](https://github.com/NovaViper/Wallpapers) to store my wallpapers and just referred to it as a flake input.

depeo:
Same, i don't know how good it is to use the impure flag..?

mio-19:
non existence path was allowed before this pr. this change helped me to discover the problem in my configuration

HeitorAugustoLN:
@axgfn if you are using flakes, just include the wallpaper in the repo, it will be automatically in the nix store

axgfn:
Wallpapers are commonly private data, though, which would necessitate configuring nix to authenticate with a private repository just for the wallpaper. Can't even encrypt the wallpaper with something like agenix because the final path would still not be in the nix store.

HeitorAugustoLN:
> Wallpapers are commonly private data, though, which would necessitate configuring nix to authenticate with a private repository just for the wallpaper. Can't even encrypt the wallpaper with something like agenix because the final path would still not be in the nix store.

I don't really consider wallpapers to be private data, but that use case is valid, so I am reverting this PR

axgfn:
Thank you, my config works again! Is there maybe some another solution that could make the option accept either a raw string or a nix path to better support both use cases?

HeitorAugustoLN:
> Thank you, my config works again! Is there maybe some another solution that could make the option accept either a raw string or a nix path to better support both use cases?

Yes, I know of a way, I will make a PR for it some time

PR #531, opened 2025-09-20
fix(workspace): make enableMiddleClickPaste option work

oidro:
This pull request fixes the `programs.plasma.workspace.enableMiddleClickPaste` option, which currently does nothing.

---

For the curious, it does nothing because `mkIf` produces an attrset that looks like this:
```nix
{
  _type = "if";
  condition = foo;
  content = bar;
}
```
using the `//` update operator on this attrset does not have the intended effect:
```nix
{
  _type = "if";
  condition = foo;
  content = bar;
  MyNewAttribute = baz;
}
```
This results in `MyNewAttribute` (or, in the case of this pull request, `Wayland.EnablePrimarySelection`) getting completely ignored.

### DIFF
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -365,14 +365,15 @@ in
                 Theme = cfg.workspace.splashScreen.theme;
               }
             );
-            kwinrc =
+            kwinrc = lib.mkMerge [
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN (approved):
Thanks for the contribution!

PR #534, opened 2025-10-03
Added --compact arg to rc2nix.py for outputting nested Nix attr sets

c0d3m0nky:
For me the way rc2nix.py outputs it's Nix conversion is unreadable. Not a criticism, just an accommodation I need, so I made it an option

When executing rc2nix.py with no arguments the output remains unchanged

```nix
{
  programs.plasma = {
    enable = true;
    shortcuts = {
      ActivityManager.switch-to-activity-1595ee1f-09c7-4e54-9969-1f629be88458 = [ ];
      "KDE Keyboard Layout Switcher"."Switch to Last-Used Keyboard Layout" = "Meta+Alt+L";
      "KDE Keyboard Layout Switcher"."Switch to Next Keyboard Layout" = "Meta+Alt+K";
      ksmserver."Lock Session" = [ "Meta+L" "Screensaver,Meta+L" "Screensaver,Lock Session" ];
      ...
    };
    ...
}
```

But with the -n|--nested-format argument it will nest the output, attr sets and lists with a single child remain in-line
```nix
{
  programs.plasma = {
    enable = true;
    shortcuts = {
      ActivityManager.switch-to-activity-1595ee1f-09c7-4e54-9969-1f629be88458 = [ ];
      "KDE Keyboard Layout Switcher" = {
        "Switch to Last-Used Keyboard Layout" = "Meta+Alt+L";
        "Switch to Next Keyboard Layout" = "Meta+Alt+K";
      };
      ksmserver."Lock Session" = [
        "Meta+L"
        "Screensaver,Meta+L"
        "Screensaver,Lock Session"
      ];
      ...
    };
    ...
}
```

I also introduced a minor change that will only quote keys when necessary using this regular expression

```^[a-z_][\w_-]*?$```

If the key begins with a letter or underscore, and the rest of the characters are all alphanumeric, underscore, or dash, it will not unnecessarily quote the key. I'm actually not 100% sure those are all the valid characters, but by making it inclusive it is fail-safe

Additional bugfixes:
- When splitting shortcut values, a regular expression pattern was being used but called in a way it was treating it as a literal string. Using the split function from re now. See [this comment](https://github.com/nix-community/plasma-manager/pull/534#issuecomment-3367374746) for an example

### DIFF
--- a/script/rc2nix.py
+++ b/script/rc2nix.py
@@ -26,7 +26,8 @@ XDG_CONFIG_HOME: str = os.path.expanduser(os.getenv("XDG_CONFIG_HOME", "~/.confi
 XDG_DATA_HOME: str = os.path.expanduser(os.getenv("XDG_DATA_HOME", "~/.local/share"))
 indent_str = '  '
 
-nix_key_valid_chars = re.compile(r'^[a-z_][\w_-]*?$', flags=re.IGNORECASE)
+rx_nix_key_valid_chars = re.compile(r'^[a-z_][\w_-]*?$', flags=re.IGNORECASE)
+rx_shortcut_value_split = re.compile(r"(?<!\\),")
 
 
 class Rc2Nix:
@@ -284,8 +285,7 @@ class Rc2Nix:
 
                 for action in actions:
                     keys = (
-                        groups[group][action]
-                        .split(r"(?<!\\),")[0]
+                        rx_shortcut_value_split.split(groups[group][action])[0]
                         .replace(r"\?", ",")
                         .replace(r"\t", "\t")
                         .split("\t")
@@ -317,7 +317,7 @@ class Rc2Nix:
 
 
 def nix_key(key: str) -> str:
-    if nix_key_valid_chars.match(key):
+    if rx_nix_key_valid_chars.match(key):
         return key
     else:
         return f'"{key}"'
--- a/script/rc2nix.py
+++ b/script/rc2nix.py
@@ -36,6 +36,15 @@ def escape_key(key: str) -> str:
         return f'"{key}"'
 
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by AlexNabokikh** [comment]
A very much needed change!

**Comment by c0d3m0nky** [comment]
I'm starting to think 'compact' is the wrong term here. Should I change it to 'pretty'? Other suggestions welcome

**Comment by AlexNabokikh** [comment]
@c0d3m0nky `--format` or `--fmt` also could work

**Comment by c0d3m0nky** [comment]
> @c0d3m0nky `--format` or `--fmt` also could work

'format' feels too vague

The other option I'm thinking is --nested

Why is coming up with good names always the hardest part? 🥲

**Comment by AlexNabokikh** [comment]
The official [Nix formatting RFC](https://github.com/NixOS/rfcs/blob/master/rfcs/0166-nix-formatting.md) uses `expanded` and `nested`attribute terms, so maybe stick to them. Adding `format` for clarity wouldn't hurt, though (e.g., `--expanded-format` or `--nested-format`)

**Comment by c0d3m0nky** [comment]
Changed it to -n|--nested-format

**Comment by c0d3m0nky** [comment]
So I found a bug in rc2nix.py but I'm not sure I should include it in this PR or wait and make a followup PR.

For shortcuts when multiple or none are set it outputs invalid values like

```nix
{
        "Window Minimize" = [
          "Meta+Down"
          "Meta+PgDown,Meta+PgDown,Minimize Window"
        ];
        "Window Move" = "none,,Move Window";
}
```
I have confirmed it happens with the HEAD of this repo too, I did not accidentally introduce this.

Should I include the fix in this PR?

c0d3m0nky:
Turned out to be an extremely simple fix, so I just included it.

magnouvean (approved):
Very cool 👍

PR #542, opened 2025-11-01
Kwin: add Blur Saturation option from Kwin 6.5.1

amenbr1:
Kwin 6.5.1 has changed the blur effect.

We can now use
```
[Effect-blur]
Saturation=3
```
to change the blur saturation.

This pull request simply adds this as an option.

### DIFF
--- a/modules/kwin.nix
+++ b/modules/kwin.nix
@@ -265,12 +265,12 @@ in
           default = null;
           example = 8;
           description = "Adds noise to the blur effect.";
+        };
         saturation = lib.mkOption {
           type = with lib.types; nullOr (ints.between 0 10);
           default = null;
           example = 1;
           description = "Adds saturation to the blur effect.";
-        };
         };
       };
       snapHelper.enable = lib.mkOption {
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN:
Thanks for contributing, have you tested it? I haven't been using plasma for a while, so I can't test it for now

amenbr1:
Yes it is tested and working on the latest plasma 6 install from unstable.

amenbr1:
Have done some more testing and it looks like the saturation can be set much higher. But this quickly becomes very very ugly. I will leave it to you to choose a sensible max value. I would personally recommend 10.

HeitorAugustoLN:
Well, looks like this option has been removed on the master branch, not sure if this should be added

HeitorAugustoLN:
https://invent.kde.org/plasma/kwin/-/commit/9bc9b2ea1a0ab2e3ecb7698e3d1c28e18914873e

amenbr1:
Ah, It seems they are changing blur again for 6.5.2. I missed that.
This should not be added.

I may create a new pull request once the changes seem more final.
Thanks for your time.

PR #548, opened 2025-11-11
escape string values passed to writeConfig for widgets

icewind1991:
allows setting values that include things like double quotes

uses `toJSON` to take care of the escaping


magnouvean (approved):
Elegant!

Issue #543, opened 2025-11-04
How to set "Application Style"?


allumik:
Running a typical classical windows style here, and always turn the MS Windows 9x on. Without getting into an obviously endless rabbit hole of styling conventions (again), what's the easiest way to set the Colors & Themes/Application Style to "MS Windows 9x"? 

I can't seem to find anything about it in the options. Or is it defined through GTK settings?

HeitorAugustoLN:
https://nix-community.github.io/plasma-manager/options.xhtml#opt-programs.plasma.workspace.windowDecorations.library

Probably this I think

allumik:
Found that the GUI settings app sets the `widgetStyle` variable in `.config/kdeglobals` file under `[KDE]`. To get the Windows 9x it is set simply to "Windows".

But the question still remains, and can't quickly find any thing about `widgetStyle` too in the docs nor issues.

Additionally, setting windowDecorations does not work for me, have to use the lookAndFeel instead.

allumik:
So it turns out it wasn't a deep hole afterall, you just need to make an issue after trying to get it to work for an hour and then the solution presents itself right after you start taking up the maintainer's valuable time...

Setting the `programs.plasma.configFile."kdeglobals".KDE.widgetStyle = "Windows"` did it for me. 

I do wonder if this should be included in the workspace attribute, as it is in the settings GUI located in near other similar settings?

ssalggnikool:
> I do wonder if this should be included in the workspace attribute

This should really be less hidden and be added there or at least documented, I agree. Something like `workspace.widgetStyle` could do.

Personally wasted around 5 hours of my life figuring out how to do this and finally found this issue

PR #554, opened 2025-11-19
feat(workspace): add widgetStyle option

ssalggnikool:
Adds `workspace.widgetStyle` to set the application/widget style as seen in system settings

Suggested in #543


magnouvean (approved):
Cool, thanks :)

PR #555, opened 2025-11-20
shortcuts: do not escape commas in service shortcuts

bmrips:
### Description

Currently, commas in the keys of shortcuts are escaped:
```
[kwin]
Show Desktop=Alt+\,

[services][foo.desktop]
_launch=Meta+\,
```
However, Plasma does not pick up the `foo.desktop` launch shortcut since it is malformed. By setting the shortcut through the UI, I found out that commas do not have to be escaped for shortcuts that are grouped under the `[services]` section, i.e. the corrected example would be
```
[kwin]
Show Desktop=Alt+\,

[services][foo.desktop]
_launch=Meta+,
```

This PR fixes this bug.

### Tests

I successfully tested this change on my own system.

### DIFF
--- a/modules/shortcuts.nix
+++ b/modules/shortcuts.nix
@@ -93,7 +93,6 @@
       mkGlobalShortcutFor =
         group: _action: skey:
         let
-          # Keys are expected to be a list:
           keys =
             if !builtins.isList skey then
               [ skey ]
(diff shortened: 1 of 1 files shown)


bmrips:
The pipeline now fails despite not failing before, so it is presumably already broken on `trunk`.

bmrips:
This PR is a descendant of #634.

PR #561, opened 2025-12-20
feat: declaratively install KDE cursor theme options

Asqiir:
This PR adds the option to declaratively install KDE cursor themes.

(I believe that issue has come up in the past; but I can't seem to find that again now.)

If this is accepted I can add a similar feature for all the other places where KDE downloads themes from the store.

----

Further, I figured out how to download stuff from the KDE store, the following instructions should go to the wiki or something:

```
        Instructions on how to find the download link to content from the KDE store:
        1) Browse api.kde-look.org/ocs/v1/content/categories in order to find the numerical ID of the `cursor` category.
        2) Use the API search api.kde-look.org/ocs/v1/content/data?categories=<numeric id of category>&search=<searchterm> . The entries here contain real (== without timeouts or challanges) download links. Note that the supplied MD5 sums seem not to work with `fetchzip` for some reason.

        I heavily recommend finding the exact name of what you search for in the regular (visual, non-API) KDE store first, as browsing is a lot easier there.
```

### DIFF
--- a/modules/workspace.nix
+++ b/modules/workspace.nix
@@ -168,6 +168,27 @@ in
       '';
     };
 
+    installedCursorThemes = lib.mkOption {
+      type = with lib.types; attrsOf package;
+      default = {};
+      example = {
+        "oreo_red_cursors" = pkgs.fetchzip {
+          name = "oreo_red_cursors";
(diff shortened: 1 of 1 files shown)


Asqiir:
I have further worked on this and I have found that the KDE API links frequently change.

I have fixed the code and example s.t. it takes

* pkgs.fetchzip as an argument
* and local files

Asqiir:
hi @HeitorAugustoLN 
any opinions on whether we can merge this?

Asqiir:
I think the test is already broken on the trunk.

PR #564, opened 2025-12-26
konsole: allow defining custom colorschemes with Nix values

folliehiyuki:
This approach allows further override of useful values like `Blur` and `Achor`.

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -2,6 +2,9 @@
 let
   dontCheckModules = {
     _module.check = false;
+    _module.args = {
+      inherit pkgs lib;
+    };
   };
   modules = [
     ../modules
--- a/modules/apps/konsole.nix
+++ b/modules/apps/konsole.nix
@@ -1,4 +1,9 @@
-{ config, lib, ... }:
+{
+  config,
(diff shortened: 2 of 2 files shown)


bmrips:
I like this change.

What do you think about a more fine-grained approach like [this one](https://github.com/bmrips/dotfiles/blob/c87baf8ba1ddda0bbcf26d6d79255711ded383cf/home-manager/modules/programs/konsole.nix#L142)? This would yield colour names more descriptive than `Color0Intense` while still allowing to set arbitrary attributes via `extraSettings`.

folliehiyuki:
I'd like to keep the module simple, so we don't need to keep up with any schema changes of the file from Konsole in the future. I can add an example config to the option so users reading the documentation have an idea how to create a custom colorscheme right away.

magnouvean:
Yeah I probably agree that this is the right way of doing this. If you could add an example though then it would be great! Otherwise looks fine :+1:

folliehiyuki:
Example string added for the colorscheme option.

magnouvean:
Looks good!

PR #573, opened 2026-02-10
Fix possible values for vendorId

RealHypnoticOcelot:
The default value for `programs.plasma.input.touchpads.*.vendorId` is null, but the allowed type is only str. This fixes the allowed type, which allows vendorId to be left undeclared when modifying other aspects of `programs.plasma.input.touchpads.*`.

### DIFF
--- a/modules/input.nix
+++ b/modules/input.nix
@@ -88,7 +88,7 @@ let
         '';
       };
       vendorId = lib.mkOption {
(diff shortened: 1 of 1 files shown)


HeitorAugustoLN (approved):

HeitorAugustoLN:
Thanks!

PR #574, opened 2026-02-10
Revert previous patch

RealHypnoticOcelot:
Sorry, I totally had it wrong!

`touchPadToConfig` uses `vendorId`, alongside `name` and `productId`, in order to merge into kcminputrc. I don't believe it's possible to set touchpad properties without specifying these values. Sorry!


HeitorAugustoLN:
Thanks!

PR #581, opened 2026-03-11
disable startup notify for hotkeys

oldshensheep:
StartupNotify will show a bouncing terminal icon at the mouse position. It is disabled by default for commands created in the system settings, so it makes sense to disable it here.

.desktop file created by system settings
```ini
[Desktop Entry]
Exec=kitten quick-access-terminal
Name=
NoDisplay=true
StartupNotify=false
Type=Application
X-KDE-GlobalAccel-CommandShortcut=true
```


magnouvean (approved):
Nice spot!

PR #599, opened 2026-04-24
feat(shortcuts): add per-app shortcut schemes

bmrips:
This PR mainly adds support per-app shortcut schemes. I have successfully been using the per-app shortcut schemes for almost a year on my systems and want to upstream it know.

While implementing per-app shortcut schemes, I also

- simplified the shortcuts module,
- extended the documentation for `programs.plasma.shortcuts`, and
- resolved a todo in the Konsole module that mentioned to explicitly require the user setting a font in their profile definition.

### DIFF
--- a/modules/apps/konsole.nix
+++ b/modules/apps/konsole.nix
@@ -50,14 +50,6 @@ let
       font = {
         name = lib.mkOption {
           type = lib.types.str;
-          /*
-            TODO: Set default to null after adding an assertion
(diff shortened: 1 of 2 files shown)


magnouvean (approved):
Good stuff!

NicoWeio:
This PR (specifically 18eae9075ed426b615d41245cf70853786490c9d) removing `default = "Hack"` kept me from updating and made me think upstream broke something. Shouldn't there be a deprecation notice?

PR #618, opened 2026-06-04
fix(ci,example): set the file system type and disable polonium

bmrips:
The CI pipeline fails since,
- as of NixOS 26.05, `fileSystems.*.fsType` has no default value, and
- polonium has been removed from nixpkgs since Plasma 5 reached EOL.

Hence, set the file system type to `ext4` and disable polonium.


magnouvean (approved):
Thanks!

Issue #515, opened 2025-08-11
Configuration struggles, mostly regarding widgets


beaver700nh:
Hi, just discovered plasma-manager here but have had a rather frustrating time porting my KDE configuration over. I used rc2nix to save my settings and then switched my configs over to using plasma-manager. I (rather naively, perhaps) set overrideConfig to true and it wiped all my settings, which was intentional since I planned to reconfigure everything anyways. 

## Issue 1: rc2nix does not pick up desktop widgets

However, it wiped my desktop widgets too and when I tried to restore those using the rc2nix output, they were nowhere to be found. Luckily I backed up the plasma-org.kde.plasma.desktop-appletsrc file separately but for separate reasons I still wasn't able to restore it.

## Issue 2: lacking documentation

I am in the process of writing a new declarative configuration for my widgets and I've managed to put something together:
```nix
	programs.plasma = {
		enable = true;
		overrideConfig = true;
		immutableByDefault = true;

		desktop = {
			widgets = [
				{
					systemMonitor = {
						title = "CPU";
						showLegend = false;
						displayStyle = "org.kde.ksysguard.barchart";
						position = {
							horizontal = 12;
							vertical = 12;
						};
						size = {
							width = 228;
							height = 108;
						};
						sensors = [
							{
								label = "CPU 0";
								name = "cpu/cpu0/usage";
								color = "235,90,71";
							}
							# ...
							{
								label = "CPU 23";
								name = "cpu/cpu23/usage";
								color = "235,90,71";
							}
						];
					};
				}
			];
		};
	};
```
but the [documentation](https://nix-community.github.io/plasma-manager/options.xhtml) does not specify all the options for widgets so I have had to look at the source code instead [(1)](https://github.com/nix-community/plasma-manager/blob/trunk/modules/widgets/default.nix) [(2)](https://github.com/nix-community/plasma-manager/blob/trunk/modules/widgets/system-monitor.nix).

**_Specifically, one of my issues is that I can't figure out how to get the refresh interval to work for system monitor. I saw this in the source code:_**
```nix
      settings = mkOption {
        type = configValueType;
        default = null;
        description = "Extra configuration options for the widget.";
        apply = settings: if settings == null then { } else settings;
      };
```
**_but I can't find any documentation explaining the format of `settings`. This issue is bolded and italicized because it seems to be low hanging fruit that someone could hopefully help me with quickly?_**

## Issue 3: widgets don't load without workaround

[...]

PR #628, opened 2026-06-28
Per-screen desktop widget support

gorfey:
This adds a per-widget option to control which screen it renders on (requested previously as part of https://github.com/nix-community/plasma-manager/issues/515#issue-3308268191)

I am not very familiar with the codebase or the Nix language, so I'm not perfectly confident in the approach. Please let me know if there are improvements to be made.


magnouvean (approved):
Looks good, thanks!

PR #634, opened 2026-07-08
Fix CI and evaluation errors

bmrips:
CI and config evaluation throw various errors which this PR fixes:

- chore(nix-render-docs): `--toc-depth` and `--section-toc-depth` have been renamed to `--sidebar-depth`.
- chore(ci): replace `pkgs.nixfmt-rfc-style` by `pkgs.nixfmt` since the RFC style was finalized and adopted.
- refactor(okular):
  - combine `config` attributes to enable local, shared let-bindings and to require a single `lib.mkIf cfg.enable` guard only.
  - use `with lib.types;` locally since usage of `with <attrs>;` in a big scope is discouraged.
  - float `getIndexFromEnum` to its use site
- fix(okular): handle `general.showMenuBar == null`; otherwise evaluation of the Okular module fails.

### DIFF
--- a/docs/plasma-manager-options.nix
+++ b/docs/plasma-manager-options.nix
@@ -43,8 +43,7 @@ stdenv.mkDerivation {
       --style style.css \
       --script highlightjs/highlight.pack.js \
       --script highlightjs/loader.js \
-      --toc-depth 1 \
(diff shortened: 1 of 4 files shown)


magnouvean (approved):
Cool!

PR #638, opened 2026-07-11
konsole: fix passing store path to customColorSchemes

scarlet-storm:
Regression from #564 passing store paths for
programs.konsole.customColorSchemes no longer works. builtins.isPath is only true for primitive types and not string interpolated nix store paths (eg. "${pkgs.foo}/bar.colorscheme"). this causes iniFormat.generate to be called for store paths. Instead use lib.types.path.check which uses the same check as lib.types.path (primitive path or strings starting with /)

### DIFF
--- a/modules/apps/konsole.nix
+++ b/modules/apps/konsole.nix
@@ -239,7 +239,10 @@ in
         name: value:
         lib.attrsets.nameValuePair "konsole/${name}.colorscheme" {
           source =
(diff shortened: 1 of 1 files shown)


SigmaSquadron (approved):
Was just about to open a PR for the same thing. This is the correct implementation, as `isPath` does not check for the same things as `lib.types.path`.

